PR #6, opened 2016-05-11
Required modules names fix.

rokasmi:
Fixed required modules names.


etimberg:
Does the gulpfile need an update too?

rokasmi:
im not sure about gulp. Havent tested

PR #18, opened 2016-06-29
add bower.json

lovanwubing:


zachpanz88:
I agree with adding bower support, but many of the attributes of this file do not apply to this project. For example, name, description, and repository along with many other attributes are not correct.

compwright:
A better practice is to move toward npm for front-end package management, but if you can't do that there is [https://github.com/mjeanroy/bower-npm-resolver](bower-npm-resolver).

Issue #21, opened 2016-07-11
Zoom to Mouse Cursor


lysaght:
Is there any plan to implement Zoom to Mouse Cursor functionality?  

Currently zoom, zooms to middle of plot.

The plugin is passing an argument named 'center' to the zoom functions which could be used to pass mouse pointer coordinates and calculate axis min and max etc which will zoom to cursor.

etimberg:
@lysaght this is on my todo list. I think most of the functions that do the zooming already get passed the center point of the zoom, it just needs to be used.

I'm happy to merge a PR adding this

lysaght:
I raised a PR.

PR #22, opened 2016-07-16
Zoom to Mouse Cursor and Only pan deltaY on vertical scale fix

lysaght:
Zoom to Mouse Cursor (Draft) as suggested in Issue #21

### DIFF
--- a/src/chart.zoom.js
+++ b/src/chart.zoom.js
@@ -192,7 +192,7 @@ var zoomPlugin = {
 			var center = {
 				x : offsetX,
 				y : offsetY
-			}
+			};
 
 			if (e.deltaY < 0) {
 				doZoom(chartInstance, 1.1, center);
--- a/src/chart.zoom.js
+++ b/src/chart.zoom.js
@@ -41,32 +41,36 @@ function zoomIndexScale(scale, zoom, center) {
 
 }
 
-
 function zoomTimeScale(scale, zoom, center) {
 	var options = scale.options;
-	var minDelta, maxDelta;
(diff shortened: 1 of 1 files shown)


src/chart.zoom.js
```diff
@@ -41,32 +41,36 @@ function zoomIndexScale(scale, zoom, center) {
 
 }
 
-
 function zoomTimeScale(scale, zoom, center) {
 	var options = scale.options;
-	var minDelta, maxDelta;
-	var newDiff;
+	var range = scale.right - scale.left;
```
etimberg: This should account for the vertical case too

etimberg:
Code looks good. Will test this out as well.

lysaght:
Also added fix for Issue #23  (Vertical Pan pans a little horizontally as well).

src/chart.zoom.js
lysaght: Does it need to? http://www.chartjs.org/docs/#scales-time-scale says Time Scale can only be placed on the X axis.

etimberg: I think that will change when v2.2.0 of Chart.js is released. I merged a PR that adds support for the category and time axes on the left/right

lysaght: OK.  I've added a draft implementation for Time scale 'Zoom to Cursor' on the Vertical.  Wasn't able to test though as I couldn't work out how to configure a Chart.js example with Time scale on the Vertical.

etimberg:
Thanks @lysaght 
great work!

Issue #20, opened 2016-07-10
Reversable scrolling


tylerecouture:
Thank you for this plugin!  Can I request reversable scrolling?  For example one scroll click up then one scroll click down returns you to the original zoom (1:1). This might make for a more intuitive default also.

Also, my chart is on a large page and scrolling caused the page to also scroll, this worked beautifully to prevent that:
http://stackoverflow.com/questions/1459676/prevent-scroll-bubbling-from-element-to-window#answer-4593315.

courchef:
Thanks for the tip on page scrolling. I have the same issue.

I modified the wheel handler in Chart.Zoom.js at line 195 (as of version 0.1.3) with the following lines:
```
var wheelHandler = function(e) {
    e.preventDefault();
    if (e.deltaY < 0) {
        doZoom(chartInstance, 1.1);
    } else {
        doZoom(chartInstance, 0.909);
    }
};
```

It prevents a long page from scrolling when the mouse wheel is activated over the chart.
I'll do a pull requests.

etimberg:
@courchef i'm happy to merge a PR for this

etimberg:
Fixed in #24

PR #24, opened 2016-07-20
Prevent page scrolling when mouse wheel is triggered over the chart

courchef:
As mentioned in issue #20


xlammer:
hello!
` e.preventDefault();` is missing in master

PR #28, opened 2016-07-29
Add drag zoom

pml984:


src/chart.zoom.js
```diff
@@ -174,6 +174,18 @@ function positionInChartArea(chartInstance, position) {
 			(position.y >= chartInstance.chartArea.top && position.y <= chartInstance.chartArea.bottom);
 }
 
+function getYAxis(chartInstance) {
```
etimberg: this only gets the first Y axis. Is this intended?

samples/zoom-time.html
```diff
@@ -106,12 +111,13 @@
 						}
 					}]
 				},
-				pan: {
-                    enabled: true,
-                    mode: 'xy'
-                },
+				// pan: {
+    //                 enabled: true,
```
etimberg: Please remove commented code

src/chart.zoom.js
```diff
@@ -186,30 +198,97 @@ zoomNS.panFunctions.logarithmic = panNumericalScale;
 
 // Chartjs Zoom Plugin
 var zoomPlugin = {
+	afterInit: function(chartInstance) {
+		chartInstance.resetZoom = function() {
```
etimberg: It'd be nice if the `resetZoom` function could store any settings that the user had before hand. Then if the user had options at chart creation, this function could put them back rather than delete them.

etimberg:
Looks pretty good. I added some comments but haven't had a chance to test it

pml984:
I made the changes you suggested. Let me know if how I merged the scale options was what you had in mind.

pml984:
Not sure if you had a chance to look at the code any further? I would be glad to fix any issues you find.

etimberg:
@pml984 I will test this tonight and merge it :)

AhmedBHameed:
Hi @etimberg,
can you please provide us with an example of line chart using zoom range ? or at least some configuration ?

AnwarShahriar:
@etimberg is there any example for this feature?

PR #33, opened 2016-09-14
Category pan

QuantumBob:
Hi,

We recently messaged on slack.
Please consider these changes for the implementation of the category panning.
And please excuse the huge number of commits and the somewhat colourful messages.

Like I said in slack I am new to git, but now realize why you should think about your commit statements!


cmd-chart-zoom.bat
```diff
@@ -0,0 +1 @@
+start cmd /k "cd c:/xampp/htdocs/Chart.Zoom"
```
etimberg: should probably not include this file in the main repo

samples/zoom.html
```diff
@@ -132,7 +132,7 @@
                         mode: 'xy'
                     },
                     zoom: {
-                        enabled: true,
+                        enabled: false,
```
etimberg: we shouldn't disable zoom in the zoom sample

samples/zoom-bar.html
```diff
@@ -76,10 +80,12 @@
...
                     pan: {
                         enabled: true,
-                        mode: 'y'
+                        mode: 'x',
+                        speed: 10,
+                        threshold: 10
                     },
                     zoom: {
-                        enabled: true,
+                        enabled: false,
```
etimberg: This file is `zoom-bar.html`. I think it would be best to create a new pan only bar sample `pan-bar.html`

src/chart.zoom.js
```diff
@@ -114,15 +118,24 @@ function doZoom(chartInstance, zoom, center) {
...
 
-function panIndexScale(scale, delta) {
-	/*var options = scale.options;
+function panIndexScale(scale, delta, options) {
 	var labels = scale.chart.data.labels;
 	var lastLabelIndex = labels.length - 1;
+    var offsetAmt = Math.max((scale.ticks.length - ((scale.options.gridLines.offsetGridLines) ? 0 : 1)), 1);
+    var panSpeed = options.pan.speed;
+
+    console.log('panSpeed: ' + panSpeed);
```
etimberg: please remove logging

src/chart.zoom.js
```diff
@@ -145,10 +158,10 @@ function panNumericalScale(scale, delta) {
 	}
 }
 
-function panScale(scale, delta) {
+function panScale(scale, delta, panSpeed) {
```
etimberg: `panSpeed` should be `options`

src/chart.zoom.js
```diff
@@ -159,7 +172,7 @@ function doPan(chartInstance, deltaX, deltaY) {
 
 		helpers.each(chartInstance.scales, function(scale, id) {
 			if (scale.isHorizontal() && directionEnabled(panMode, 'x') && deltaX !== 0) {
-				panScale(scale, deltaX);
+				panScale(scale, deltaX, helpers.getValueOrDefault(chartInstance.options, defaultOptions));
```
etimberg: I think its best to either pass `chartInstance.options` or `panOptions` here depending on if we want the pan method to have access to all the options or just the pan options. It might make sense to add a check further up that ensures the options are present so that we don't have to have all kinds of these checks everywhere (i'm fine with doing that in a separate pull request)

src/chart.zoom.js
```diff
@@ -343,6 +357,7 @@ var zoomPlugin = {
 			mc.on('panend', function(e) {
 				currentDeltaX = null;
 				currentDeltaY = null;
+                zoomNS.panCumulativeDelta = 0;
```
etimberg: looks like the spacing is different here according to GH

QuantumBob: Not sure why. My copy looks lined up. Have deleted all indents and reset it with tabs

src/chart.zoom.js
QuantumBob: You had already done it!

`function doPan(chartInstance, deltaX, deltaY) {`
`var panOptions = chartInstance.options.pan;`
`if (panOptions && helpers.getValueOrDefault(panOptions.enabled, defaultOptions.pan.enabled)) {`
		`var panMode = helpers.getValueOrDefault(chartInstance.options.pan.mode, defaultOptions.pan.mode);`

QuantumBob:
Hi,

I've done all the changes requested above. 

You had already got the panOptions variable check in doPan.
Most of the other mistakes were silly 11pm shouldn't haves.
There is now a bar-pan.html sample with the speed option set in the options.

QuantumBob:
The indents in panIndexScale() don't match up with the rest.
It is annoying me but I haven't been able to fix it

src/chart.zoom.js
QuantumBob: Noticed that this wasn't picking up the default pan speed if there wasn't one in the user options.
doPan() now sets panOptions.speed  to either the user or default value before passing panOptions to panScale

PR #39, opened 2016-09-20
Category zoom

QuantumBob:
Hi,

I've got the category zooming to work.
Added an option under zoom to change how fast it zooms called sensitivity (we can rename if you like!)
Added a check that options.sensitivity exists or uses default.
Had to use another global under the zoomNS namespace called zoomCumulativeDelta.
Zooms from left or right then fills out the other side.
New sample added, but we could just use the existing zoom-bar or rename if you want.

Thanks

Rob


etimberg:
This works quite well 😄 
I did notice a bug in Chart.js itself where the bar width is incorrectly calculated when the x axis is zoomed.

The last commit says 'change indents to match github' but I get a ton of diffs. Is that intended?

src/chart.zoom.js
```diff
@@ -19,185 +19,222 @@ var panFunctions = zoomNS.panFunctions = zoomNS.panFunctions || {};
...
+                        } else if(center.x > scale.width/2){
+                                if (maxIndex >= lastLabelIndex){
+                                        minIndex = Math.max(0, minIndex - 1);
+                                } else{
+                                        maxIndex = Math.min(lastLabelIndex, maxIndex + 1);
+                                }
+                        }
+                        zoomNS.zoomCumulativeDelta = 0;
+                }
+                if(zoomNS.zoomCumulativeDelta > 0){
```
etimberg: Minor point:
I think this should be `else if (zoomNS.zoomCumulativeDelta > 0)` unless I've missed a way that the cumulative delta can become negative in the first block

src/chart.zoom.js
```diff
@@ -19,185 +19,222 @@ var panFunctions = zoomNS.panFunctions = zoomNS.panFunctions || {};
...
+        var minIndex = scale.minIndex;
+        var lastLabelIndex = labels.length - 1;
+        var maxIndex = scale.maxIndex;
+        var sensitivity = zoomOptions.sensitivity;
+
+        zoomNS.zoomCumulativeDelta = zoom > 1 ? zoomNS.zoomCumulativeDelta + 1 : zoomNS.zoomCumulativeDelta - 1;
+
+        if (Math.abs(zoomNS.zoomCumulativeDelta) > sensitivity){
+                if(zoomNS.zoomCumulativeDelta < 0){
+                        if(center.x <= scale.width/2){
```
etimberg: I don't think this check is entirely correct. It only works if `scale.left === 0`.

I think it would be better as
`if (center.x <= scale.left + (scale.width / 2))`

src/chart.zoom.js
```diff
@@ -19,185 +19,222 @@ var panFunctions = zoomNS.panFunctions = zoomNS.panFunctions || {};
...
+                                if (minIndex <= 0){
+                                        maxIndex = Math.min(lastLabelIndex, maxIndex + 1);
+                                } else{
+                                        minIndex = Math.max(0, minIndex - 1);
+                                }
+                        } else if(center.x > scale.width/2){
```
etimberg: see above comment

src/chart.zoom.js
```diff
@@ -19,185 +19,222 @@ var panFunctions = zoomNS.panFunctions = zoomNS.panFunctions || {};
...
+                                if (maxIndex >= lastLabelIndex){
+                                        minIndex = Math.max(0, minIndex - 1);
+                                } else{
+                                        maxIndex = Math.min(lastLabelIndex, maxIndex + 1);
+                                }
+                        }
+                        zoomNS.zoomCumulativeDelta = 0;
+                }
+                if(zoomNS.zoomCumulativeDelta > 0){
+                        if(center.x <= scale.width/2){
```
etimberg: see above comment

src/chart.zoom.js
```diff
@@ -19,185 +19,222 @@ var panFunctions = zoomNS.panFunctions = zoomNS.panFunctions || {};
...
+                                minIndex = minIndex < maxIndex ? minIndex = Math.min(maxIndex, minIndex + 1) : minIndex;
+                        } else if(center.x > scale.width/2){
```
etimberg: same as above

src/chart.zoom.js
etimberg: Also, since this is used 4 times, it helps for minification if we store it in a variable at the top of the function

etimberg: Another thought is that checking the x only works if the scale is horizontal. `scale.isHorizontal()` returns true if this case is met. This should be updated to work for vertical index scales such as those that can occur in a horizontal bar chart.

QuantumBob: I was trying to not leave a dead zone when the pointer/pinch centre was at dead center as the other check is just for greater than.
Hadn't thought about minification, but I will now!
I'll fix the isHorizontal issue too
The indents are still bugging me too but I didn't have time to sort then out. Next commit will hopefully. Do you prefer tab or space indents?

src/chart.zoom.js
QuantumBob: You are correct!

QuantumBob:
Internet access is awful here at the moment. Will get it done asaic

etimberg:
Awesome! Per your tabs and spaces question, i prefer tabs

PR #55, opened 2016-11-17
Export as module

### LINKED ISSUE #16: Create as an actual module, i.e. export `zoomPlugin`
I want to be able to register this plugin with my own instance of Chart, which is not named `Chart` in my case. i.e. I want:

``` js
import Chart2 from 'chart.js';
import zoomPlugin from 'chart.zoom.js';

Chart2.pluginService.register(zoomPlugin);
```

compwright:
Resolves #16


etimberg:
Is there any reason that `module` would be undefined at all?

compwright:
@etimberg not since we are building with Browserify.

etimberg:
:+1:

Issue #37, opened 2016-09-19
How can i set pan and zoom limits ?


ivanicharts:
Can I somehow set limit on min and max values on the x axis so user couldn't pan beyond the 0 ?

etimberg:
@ivanicharts this is not yet possible. I'd be happy to merge a PR that adds this functionality

ivanicharts:
@etimberg, thanks, but there are any ways I can do it by myself ?

nolanthorn:
@etimberg I have handled this (for TimeScales, at least, and it doesn't appear to me that the others are any harder), but it seems to me I need to compute the minimum and maximum x and y values from the data each time doPan is called. This is an obvious performance issue on large datasets, so my question is: can I access the minimum and maximum data values for each dataset on a property of the ChartInstance?

If so, I'm happy to try and make a PR for this, it is of fair urgency in a project I'm involved with.

etimberg:
You'd need to modify the functions inhttps://github.com/chartjs/Chart.Zoom.js/blob/master/src/chart.zoom.js such as `zoomNumericalScale` and `zoomIndexScale` to check if the zoom goes over the limits and cap it if it does

PR #74, opened 2017-05-09
Added support for setting lower/upper limit for zoom/pan range

bjornnorman:
I added two functions for enforcing the limits and ensured they are called by all zoom/pan functions.
To support both x and y directions, I supplied the detected direction with the options object before passing it along.
The solutions works excellently in my case with a horizontal timescale graph, but I haven't tested it with other scale types, but in theory it should handle that too as it is.
The README.MD file was updated to include the new range settings.
Hope you find this useful - let me know if you have any questions :)

### DIFF
--- a/src/chart.zoom.js
+++ b/src/chart.zoom.js
@@ -23,12 +23,12 @@ var defaultOptions = zoomNS.defaults = {
 		enabled: true,
 		mode: 'xy',
 		speed: 20,
-		threshold: 10,
+		threshold: 10
 	},
 	zoom: {
 		enabled: true,
 		mode: 'xy',
-		sensitivity: 3,
+		sensitivity: 3
 	}
 };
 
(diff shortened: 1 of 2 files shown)


bjornnorman:
Damnit. Didn't see you already had another pull request for this feature...
Very well, now you'll have to decide which is the more elegant! ;)

compwright:
Resolves #37

src/chart.zoom.js
```diff
@@ -42,6 +42,26 @@ function directionEnabled(mode, dir) {
...
+		var rangeMax = zoomPanOptions.rangeMax[zoomPanOptions.scaleAxes];
+        if (newMax > rangeMax) {
+            newMax = rangeMax;
+        }
+    }
+    return newMax;
+}
+
+function rangeMinLimiter(zoomPanOptions, newMin) {
+    if (zoomPanOptions.scaleAxes && zoomPanOptions.rangeMax && zoomPanOptions.rangeMax[zoomPanOptions.scaleAxes]) {
```
compwright: Appears to be typos: should it be `zoomPanOptions.rangeMin`?

bjornnorman: Good catch - you're right, I missed that line when I cut and pasted the Max function ;) I've updated the pull request now.

ericnkatz:
Is there a working example of this on Codepen somewhere? I cannot seem to get this to work.

ericnkatz:
@bjornnorman @compwright the limits don't work when values are 0 due to type coercion?

bjornnorman:
Yes, there should probably be added a "=== null" test to avoid ignoring "0" limits, and a default config value of "null" in case the max/min params aren't specified.

sridhark158:
Is there a working example of this codepen/jsfiddle ?

arnab333:
Is there any working example for "rangeMin" and "rangeMix"?

PR #79, opened 2017-05-30
Only update min/max when bot have changed. closes #78

### LINKED ISSUE #78: [Pan-Limits] "Zoom" when hitting the border
I just tested the new version 0.5.0 of this great plugin and I had the following problem:
When I pan to the right of my chart and hit the border. The chart is zooming in.

This is my pan-config:
```
pan: {
      enabled: true,
      mode: 'x',
      rangeMax: {
        // Format of max pan range depends on scale type
        x: moment('2017-05-31T00:00:00+02:00')
      }
}
```
Zoom is disabled. Type is 'time'

To my mind the problem is the update of min value without checking if max has been reached
Edited the pan-functions to only update min/max if a border was exceeded. This solved my problem. 

I will open a pull request for that. It's just my quick fix and I have no idea if this breaks some other functionality. :)

meisterlampe:
I hope this closes #78.
It's just a quick fix and for my use case it seems to work fine. My use case is:
- 'time' scale
- Zoom disabled
- Pan only on x-axis
- Panning restricted by rangeMin.x and rangeMax.x

`panNumericalScale` has also been edited the same way as `panTimeScale`. I thought, it might behave similar, but it hasn't been tested.

I have no idea if I solved it the wrong way or if something else is broken by this changes. I just want to contribute :)

### DIFF
--- a/src/chart.zoom.js
+++ b/src/chart.zoom.js
@@ -227,13 +227,13 @@ function panNumericalScale(scale, delta, panOptions) {
 	var newMax = rangeMaxLimiter(panOptions, tickOpts.max);
 	var newMin = rangeMinLimiter(panOptions, tickOpts.min);
 
-	var limitedTimeDelta = delta < 0 ? newMax - end : newMin - start;
+	var limitedNumericDelta = delta < 0 ? newMax - end : newMin - start;
 
-	if (Math.abs(limitedTimeDelta) < 0.1) // Use an epsilon factor to ignore fp rounding errors
-		limitedTimeDelta = 0;
+	if (Math.abs(limitedNumericDelta) < 0.1) // Use an epsilon factor to ignore fp rounding errors
+		limitedNumericDelta = 0;
 
-	tickOpts.max = newMax + limitedTimeDelta;
-	tickOpts.min = newMin + limitedTimeDelta;
+	tickOpts.max = newMax + limitedNumericDelta;
+	tickOpts.min = newMin + limitedNumericDelta;
 }
 
 function panScale(scale, delta, panOptions) {


src/chart.zoom.js
```diff
@@ -201,8 +201,13 @@ function panIndexScale(scale, delta, panOptions) {
 
 function panTimeScale(scale, delta, panOptions) {
 	var options = scale.options;
-	options.time.min = rangeMinLimiter(panOptions, scale.getValueForPixel(scale.getPixelForValue(scale.firstTick) - delta));
-	options.time.max = rangeMaxLimiter(panOptions, scale.getValueForPixel(scale.getPixelForValue(scale.lastTick) - delta));
+	
+	var newMax = rangeMaxLimiter(panOptions, scale.getValueForPixel(scale.getPixelForValue(scale.lastTick) - delta));
+	var newMin = rangeMinLimiter(panOptions, scale.getValueForPixel(scale.getPixelForValue(scale.firstTick) - delta));
+	if (newMax != options.time.max && newMin != options.time.min) {
```
compwright: I'm thinking these should be `<` and `>` instead of equality operators.

src/chart.zoom.js
```diff
@@ -217,8 +222,13 @@ function panNumericalScale(scale, delta, panOptions) {
 		tickOpts.min = scale.getValueForPixel(scale.getPixelForValue(start) - delta);
 		tickOpts.max = scale.getValueForPixel(scale.getPixelForValue(end) - delta);
 	}
-	tickOpts.min = rangeMinLimiter(panOptions, tickOpts.min);
-	tickOpts.max = rangeMaxLimiter(panOptions, tickOpts.max);
+	
+	var newMax = rangeMaxLimiter(panOptions, tickOpts.max);
+	var newMin = rangeMinLimiter(panOptions, tickOpts.min);
+	if (newMax != tickOpts.max && newMin != tickOpts.min) {
```
compwright: Ditto

src/chart.zoom.js
meisterlampe: I just want to make sure I got everything right. So this:
```
newMax != options.time.max && newMin != options.time.min
```
will have to change to
```
newMax > options.time.max && newMin > options.time.min
|| newMax < options.time.max && newMin < options.time.min
```
options.time.max and options.time.min seem to be undefined at first. Added this check and ended up with:
```
if (!options.time.max || !options.time.min     //Check is not working on first iteration
     || newMax > options.time.max && newMin > options.time.min
     || newMax < options.time.max && newMin < options.time.min) {
	
        [...]

}
```

Would this be ok?

compwright: I was just thinking something like:

```javascript
newMax <= options.time.max && newMin >= options.time.min
```

To address the undefined issue:

```javascript
(!options.time.max || newMax <= options.time.max) && (!options.time.min || newMin >= options.time.min)
```

I haven't tested this, FYI.

meisterlampe: Maybe I'm using a wrong approach? (checking options.time.max for undefined doesn't look that nice)
I could also check, if the calculated new max/min is different than the output of the function rangeMax/Min/Limiter(...). Only if nothing has been cut by the function, we could set options.time.max/.min.

Something like:
```
var options = scale.options;

var targetMax = scale.getValueForPixel(scale.getPixelForValue(scale.lastTick) - delta);
var newMax = rangeMaxLimiter(panOptions, targetMax);
var targetMin = scale.getValueForPixel(scale.getPixelForValue(scale.firstTick) - delta)
var newMin = rangeMinLimiter(panOptions, targetMin);
	
// Only set min/max, if target-values are equal to the result-values of the range-limiter-functions
if (newMax == targetMax && newMin == targetMin) {
	options.time.max = newMax;
	options.time.min = newMin;
}
```

What's your opinion?

meisterlampe: I tested your suggestion, but unfortunately that didn't work. It blocks panning to the left and has also the problem with the zoom effect, when hitting the right border.

Here are the cases that occur normally:

Case        | Max | Min|Update values?
----------|------|----|--------
pan left | newMax < oldMax | newMin < oldMin | yes
pan right | newMax > oldMax | newMin > oldMin | yes
pan left, rangeLimited | newMax < oldMax | newMin == oldMin | no
pan right, rangeLimited | newMax == oldMax | newMin > oldMin | no

The idea was to allow updating the values only in the first two cases.

compwright: Thanks, that's a helpful visualization.

My main worry with the equality check is the possibility of over panning or zooming. If you're going to do equality, I recommend factoring in an epsilon (fudge) factor:

```
Math.abs(min - minLimit) < epsilon
```

where `epsilon = 0.1` as a substitute for normal equality.

meisterlampe: Ah, ok.. You're talking about rounding errors when doing floating-point computations, where it's not a good idea to just check for equality, right?

This applies to `panNumericScale` (which I don't use and haven't tested O:-) ). 
I will try to refactor this one and add an epsilon factor.

But in this case `panTimeScale` is doing it's calculations using momentjs-dates. Checking for equality on them should be ok, shouldn't it?

compwright: Yes, we have to be careful comparing FPs.

Equality always fails with date objects, although other comparisons can be used. momentjs has `isSame()` for comparing dates, and you should use that.

src/chart.zoom.js
```diff
@@ -201,8 +201,14 @@ function panIndexScale(scale, delta, panOptions) {
...
-	options.time.min = rangeMinLimiter(panOptions, scale.getValueForPixel(scale.getPixelForValue(scale.firstTick) - delta));
-	options.time.max = rangeMaxLimiter(panOptions, scale.getValueForPixel(scale.getPixelForValue(scale.lastTick) - delta));
+	
+	var limitedMax = rangeMaxLimiter(panOptions, scale.getValueForPixel(scale.getPixelForValue(scale.lastTick) - delta));
+	var limitedMin = rangeMinLimiter(panOptions, scale.getValueForPixel(scale.getPixelForValue(scale.firstTick) - delta));
+
+	var limitedTimeDelta = delta < 0 ? limitedMax.diff(scale.lastTick) : limitedMin.diff(scale.firstTick);
+
+	options.time.max = scale.lastTick.add(limitedTimeDelta);
+	options.time.min = scale.firstTick.add(limitedTimeDelta);
```
meisterlampe: How about this?
My first approach had a problem with the last movement that hits the border.

For example:
lastTick: 2017-05-01T23:00:00Z, delta (time-duration):  2 hours, rangeMax: 2017-05-02T00:00:00Z.
This would result in an effective time delta of 1 hour for max, but the min value will change 2 hours.

My intention with this code was to ensure, that the same delta is added to min and max.

src/chart.zoom.js
```diff
@@ -226,10 +226,14 @@ function panNumericalScale(scale, delta, panOptions) {
...
-		tickOpts.min = newMin;
-	}
+
+	var limitedTimeDelta = delta < 0 ? newMax - end : newMin - start;
+
+	if (Math.abs(limitedTimeDelta) < 0.1) // Use an epsilon factor to ignore fp rounding errors
+		limitedTimeDelta = 0;
+
+	tickOpts.max = newMax + limitedTimeDelta;
+	tickOpts.min = newMin + limitedTimeDelta;
```
meisterlampe: Same here.
I wanted to ensure the same delta is being added to min/max.
Plus i added an epsilon factor.

src/chart.zoom.js
```diff
@@ -217,8 +223,17 @@ function panNumericalScale(scale, delta, panOptions) {
 		tickOpts.min = scale.getValueForPixel(scale.getPixelForValue(start) - delta);
 		tickOpts.max = scale.getValueForPixel(scale.getPixelForValue(end) - delta);
 	}
-	tickOpts.min = rangeMinLimiter(panOptions, tickOpts.min);
-	tickOpts.max = rangeMaxLimiter(panOptions, tickOpts.max);
+	
+	var newMax = rangeMaxLimiter(panOptions, tickOpts.max);
+	var newMin = rangeMinLimiter(panOptions, tickOpts.min);
+
+	var limitedTimeDelta = delta < 0 ? newMax - end : newMin - start;
```
compwright: This should be renamed - it isn't a time delta, it's a numeric one.

meisterlampe: You're right... fixing that :)

meisterlampe:
Hi.. is everything fine with the changes?

benmccann:
Is there a reason that `panTimeScale` and `panNumericalScale` are being changed but not `panIndexScale`? It seems like we might want to change them all for consistency

src/chart.zoom.js
```diff
@@ -217,8 +223,17 @@ function panNumericalScale(scale, delta, panOptions) {
...
 	}
-	tickOpts.min = rangeMinLimiter(panOptions, tickOpts.min);
-	tickOpts.max = rangeMaxLimiter(panOptions, tickOpts.max);
+	
+	var newMax = rangeMaxLimiter(panOptions, tickOpts.max);
+	var newMin = rangeMinLimiter(panOptions, tickOpts.min);
+
+	var limitedNumericDelta = delta < 0 ? newMax - end : newMin - start;
+
+	if (Math.abs(limitedNumericDelta) < 0.1) // Use an epsilon factor to ignore fp rounding errors
```
benmccann: We always include curly braces with our if statements

meisterlampe: This Pull Request is really old now. I actually thought, that this project is dead and deleted my forked repo. I guess, this PR cannot be updated and I have to create a new PR right?

benmccann: It was sort of dead. I just joined to help revive it and have merged a few PRs today. Yes, as far as I know, you would have to create a new PR. I'm not sure how else to do it unfortunately.

meisterlampe: Cool. Good to hear, was hoping for some rescuer like you to come :-) 👍 
I'm a bit busy right now, but will look into this in about two weeks.

meisterlampe:
New PR #119 has been created, as the original forked repo of this PR has been deleted.

PR #94, opened 2017-07-26
Fix pinch center calculation closes #93

### LINKED ISSUE #93: [Zoom] Pinch uses wrong center
The center point used for pinch isn't competely correct when the chart is not aligned left.
For `wheel` event a correction is made using `getBoundingClientRect()`.
I think just copying this code to the handlePinch-function should fix it. Will provide a pull request for that.

meisterlampe:
The center point used for pinch isn't competely correct when the chart is not aligned left.
For wheel event a correction is made using getBoundingClientRect().
I copied the code to the function `handlePinch` and it was working perfectly fine in my use case.
This should close #93 

Best wishes
MeisterLampe

### DIFF
--- a/src/chart.zoom.js
+++ b/src/chart.zoom.js
@@ -396,7 +396,15 @@ var zoomPlugin = {
 			var currentPinchScaling;
 			var handlePinch = function handlePinch(e) {
 				var diff = 1 / (currentPinchScaling) * e.scale;
-				doZoom(chartInstance, diff, e.center);
+				var rect = e.target.getBoundingClientRect();
+				var offsetX = e.center.x - rect.left;
+				var offsetY = e.center.y - rect.top;
+				var center = {
+					x : offsetX,
+					y : offsetY
+				};
+
+				doZoom(chartInstance, diff, center);
 
 				// Keep track of overall scale
 				currentPinchScaling = e.scale;


benmccann:
@Hadatko do you have thoughts about this PR? Would it be compatible with your PR https://github.com/chartjs/chartjs-plugin-zoom/pull/113?

Hadatko:
@benmccann from what i see, you will get one merge conflict and yo need be sure that after merge conflict this function call will looks like: doZoom(chartInstance, diff, center, xy);
After then i think everything will works. 

I am apologizing i can't try it, i am not working with chart.js anymore.

benmccann:
Can you provide more details about what the issue is and why this fixes it?

meisterlampe:
Sure, here is my scenario:
In my Ionic-App there is a menu on the left side. Because of this, the chart is displayed with an offset from the left. Zooming by the mouse-wheel is working perfectly fine, but on a mobile device pinching is not correct:

Expected behaviour in this case:
The chart is zooming to the center point of the pinch gesture.

What I get:
The point to which the chart is zooming has an offset to the center of my actual pinch gesture. The offset seems to be the width of my menu. (e.g. I want to zoom in to a peak in my data, but the chart is actually zooming to a point to the side of the peak)

In the code I saw, that using the mouse-wheel is working, because the actual position of the chart is beeing respected. 
To correct pinch to zoom, I therefore copied lines from the wheel-handler to the pinch handler.
In my app pinching is now working perfectly fine.

PR #99, opened 2017-08-24
Allow `0` as min/max value

lacolaco:
Currently, `rangeMin.x` and other min/max options are not accepting `0` as valid value because `0` is _falsy_ value. But in this context `0` should be allowed.

### DIFF
--- a/src/chart.zoom.js
+++ b/src/chart.zoom.js
@@ -43,7 +43,7 @@ function directionEnabled(mode, dir) {
 }
 
 function rangeMaxLimiter(zoomPanOptions, newMax) {
-	if (zoomPanOptions.scaleAxes && zoomPanOptions.rangeMax && zoomPanOptions.rangeMax[zoomPanOptions.scaleAxes]) {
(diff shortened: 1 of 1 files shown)


src/chart.zoom.js
```diff
@@ -43,7 +43,7 @@ function directionEnabled(mode, dir) {
 }
 
 function rangeMaxLimiter(zoomPanOptions, newMax) {
-	if (zoomPanOptions.scaleAxes && zoomPanOptions.rangeMax && zoomPanOptions.rangeMax[zoomPanOptions.scaleAxes]) {
+	if (zoomPanOptions.scaleAxes && zoomPanOptions.rangeMax && zoomPanOptions.rangeMax[zoomPanOptions.scaleAxes] != null) {
```
benmccann: perhaps it would be better to call `helpers.isNullOrUndef` here

benmccann:
Thank you for this. I've made this change using `isNullOrUndef` here: https://github.com/chartjs/chartjs-plugin-zoom/commit/7363ff5474252aadc19207773f4ae4b7f5945da6

PR #104, opened 2017-09-14
Stop event listening when zoom is not enabled

### LINKED ISSUE #102: Zoom affecting all charts on page
I'm running chart.js version 2.5.0, with multiple charts on one page. However, when I zoom into one chart and scroll through the page, all the other charts reflect the zooming change made in the other chart. Is there an option to disable this behavior that I didn't find in the documentation, or is this a bug?

lacolaco:
Currently, always the plugin catches events. It should be skipped when zoom is not enabled.


benmccann:
@simonbrunel @etimberg this PR seems fine to merge to me

Issue #90, opened 2017-07-19
No x-axis pan or zoom using Chart.js 2.6.0


meisterlampe:
chartjs-plugin-zoom seems to have some problems in conjunction with Chart.js 2.6.0:
In my case, I cannot pan or zoom on the x-axis.

When I simply run `npm install chart.js@2.5.0` x-axis panning and zooming starts working again immediately. Running `npm install chart.js@2.6.0` breaks x-axis pan&zoom again. So it looks like a problem with the chart.js version.

Are you experiencing the same problems or is this working for you?
Any ideas what I can do to upgrade to the latest version of chartjs-plugin-zoom (it has 2.6.0 as dependency)?
Please help.

In my config I use:

   ```
pan: {
      enabled: true,
      mode: 'xy',
      rangeMin: {
        x: null,
        y: null
      },
      rangeMax: {
        x: null,
        y: null
      }
    },
    zoom: {
      enabled: true,
      mode: 'xy',
    }
```

Enviroment:
- Ionic 3.5.3
- ng2-charts 1.6.0

Edit:
Chart-type: time

WeaVaer:
+1 having exactly the same problem.
Even the chart in the live codepen demo link doesn't work on zooming.
I would like to go to chart.js 2.5.0 but I am worried I will have to change my app code and also might be missing some features by downgrading to chart.js 2.5.0

adeel3612:
i have chartjs 2.6 zoom is working fine on x-axis. But panning of x-axis (simple x-axis which is by default category axis) is not supported. Whether it is 2.5 or any version.

WeaVaer:
My chart is a timeline series (x axis is date-time).
Neither zoom nor pan works for chart.js 2.6.
As soon as I tried @meisterlampe 's idea using chart.js 2.5 without changing any line of my code, both pan and zoom started working.
Weird...

adeel3612:
weird! i have not experienced this in my time axis.

meisterlampe:
I too have a time series with moment.js-datetime-objects on the x-Axis (will add this info to my original post, maybe it's important).

adeel3612:
yeah may be may be you should stick 2.5 chart.

meisterlampe:
`zoom-time.html` in the [samples-folder](https://github.com/chartjs/chartjs-plugin-zoom/tree/master/samples) is a good example:

[...]

PR #105, opened 2017-09-21
Fix for zoom and pan on time axis

jarrodbell:
scale.firstTick and lastTick are not longer available in latest Chart.js releases. So change to using min and max values for pan/zoom calculations.

### DIFF
--- a/src/chart.zoom.js
+++ b/src/chart.zoom.js
@@ -122,8 +122,8 @@ function zoomTimeScale(scale, zoom, center, zoomOptions) {
 	var minDelta = newDiff * min_percent;
 	var maxDelta = newDiff * max_percent;
 
-	options.time.min = rangeMinLimiter(zoomOptions, scale.getValueForPixel(scale.getPixelForValue(scale.firstTick) + minDelta));
-	options.time.max = rangeMaxLimiter(zoomOptions, scale.getValueForPixel(scale.getPixelForValue(scale.lastTick) - maxDelta));
+	options.time.min = rangeMinLimiter(zoomOptions, scale.getValueForPixel(scale.getPixelForValue(scale.min) + minDelta));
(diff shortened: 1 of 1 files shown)


jarrodbell:
This fixes #90

FeNoMeNa:
progress about this fix ?

statement2020:
+1 on this please

benmccann:
It looks like those variables were only in the time scale and were removed in https://github.com/chartjs/Chart.js/pull/3914

benmccann:
@simonbrunel @etimberg does my last comment sound right? let's go ahead and merge this PR because the plugin is currently broken without it

benmccann:
@jarrodbell the other thing we might want to do is to update the Chart.js version in package.json. It currently has `"chart.js": "^2.1.5",`, but `2.1.5` won't work after this change. We should set it to `2.6.0` instead

simonbrunel:
I thought scale min/max was around for a long time, way before 2.6?

benmccann:
It seems it wasn't used in the time scale at least. I'm not sure about the others. But here's the time scale state at the commit previous to the change I mentioned earlier and it doesn't reference `me.min` or `me.max` anywhere: https://github.com/chartjs/Chart.js/blob/9f3b51a80ce96578718267711e8b65c1ec8c25c1/src/scales/scale.time.js

In either case, the current time scale code doesn't use `firstTick` and `lastTick`, which means this code will need to be updated accordingly

simonbrunel:
sounds good!

benmccann:
Great, let's merge this then. If you'd like, I can do it and help maintain this repo if you give me access

compwright:
@etimberg you have another volunteer for a contributor!

compwright:
@benmccann thanks for pitching in and helping out with this project!

benmccann:
@compwright no problem. I could use some help figuring out how to cut a release with the fixes. I was asking on the Chart.js slack channel if you're on there

PR #119, opened 2018-01-02
Issue78 zoom when hitting border

### LINKED ISSUE #78: [Pan-Limits] "Zoom" when hitting the border

meisterlampe:
This closes #78 
Replaces PR #79 (the repo it was based on was deleted)
Maybe the function panNumericalScale has the same issue. I want to focus the actual issue, so I'm splitting this up into two PRs


meisterlampe:
Hm. There a quite some commits now for such a small change. A good candidate for squashing? :)

PR #144, opened 2018-05-25
Removed unnecessary chart updates when drag zoom is enabled

MPursche:
Hello,

this pull request removes unnecessary updates of the chart if drag zoom is used.
The mouse move callback previously updated the chart on every move event, even if no drag zoom gesture was started.
Additionally, the `_dragZoomStart` and `_dragZoomEnd` were not removed before the `doZoom` call which could result in drawing the zoom selection after the execution of the zoom.

Greetings

Marcel


chartjs-plugin-zoom.js
```diff
@@ -390,15 +388,16 @@ var zoomPlugin = {
 					var chartDistance = chartArea.right - chartArea.left;
 					var zoom = 1 + ((chartDistance - dragDistance) / chartDistance );
 
+					// Remove drag start and end before diagram update to stop drawing selected area
```
benmccann: can you change "diagram" to "chart" in this comment to match the terminology we normally use

PR #145, opened 2018-05-25
Fix reset zoom

MPursche:
Hello,

this pull request fixes the reset zoom method. The previous version that used `helpers.configMerge` did not work for some reason. The `min` and `max` options are now restored to their previous value manually.

Greetings

Marcel


chartjs-plugin-zoom.js
```diff
@@ -319,26 +319,24 @@ zoomNS.zoomCumulativeDelta = 0;
...
 	afterInit: function(chartInstance) {
-		helpers.each(chartInstance.scales, function(scale) {
-			scale.originalOptions = JSON.parse(JSON.stringify(scale.options));
+		helpers.each(chartInstance.scales, function (scale) {
+			scale.originalOptions = helpers.clone(scale.options);
 		});
 
-		chartInstance.resetZoom = function() {
-			helpers.each(chartInstance.scales, function(scale, id) {
+		chartInstance.resetZoom = function () {
```
benmccann: Can you please remove the space added after `function` here and below as well

chartjs-plugin-zoom.js
```diff
@@ -319,26 +319,24 @@ zoomNS.zoomCumulativeDelta = 0;
 // Chartjs Zoom Plugin
 var zoomPlugin = {
 	afterInit: function(chartInstance) {
-		helpers.each(chartInstance.scales, function(scale) {
-			scale.originalOptions = JSON.parse(JSON.stringify(scale.options));
+		helpers.each(chartInstance.scales, function (scale) {
```
benmccann: remove space after `function`

chartjs-plugin-zoom.js
```diff
@@ -319,26 +319,24 @@ zoomNS.zoomCumulativeDelta = 0;
...
+			scale.originalOptions = helpers.clone(scale.options);
 		});
 
 		chartInstance.resetZoom = function() {
-			helpers.each(chartInstance.scales, function(scale, id) {
+			helpers.each(chartInstance.scales, function (scale, id) {
```
benmccann: remove space after `function`

benmccann:
@MPursche this PR will need to be rebased

benmccann:
Thanks @MPursche. I'm also concerned that `configMerge` may not be working correctly upstream. I wrote a test https://github.com/chartjs/Chart.js/pull/5695 and it seems to be working. Can you give me an example of what the config looks like which doesn't merge correctly?

MPursche:
@benmccann It's been a while since I discovered this bug so I'm not quite sure anymore. But I think it was a problem with `moment.js` objects. I used a `TimeScale` on the x axis and configured time scale range with `moment.js` objects. `configMerge` copied the properties of a `moment.js` object into a plain object instead of cloning it properly(there is a `.clone()` method for moment objects). The cloned objects were missing the `moment.js` methods.

MPursche:
I was able to reproduce the bug in `configMerge`:
```javascript
var moment = require('moment');
var Chart = require('chart.js');
var helpers = Chart.helpers;

var date = moment();
var config {};
config = helpers.configMerge(config, { min: date });
console.log(config.min); // <- this is not a moment object
var newDate = moment(config.min); // <- This will result in an error 'TypeError: locale.longDateFormat is not a function'
```

benmccann:
Thanks!

src/chart.zoom.js
```diff
@@ -317,16 +317,14 @@ var zoomPlugin = {
...
-					delete timeOptions.max;
+					timeOptions.min = scale.originalOptions.time.min;
+					timeOptions.max = scale.originalOptions.time.max;
 				}
 
 				if (tickOptions) {
-					delete tickOptions.min;
-					delete tickOptions.max;
+					tickOptions.min = scale.originalOptions.ticks.min;
+					tickOptions.max = scale.originalOptions.ticks.max;
```
jannisgt: This makes only sense if the min/max values are initially set and they should always be in that range.
If you change the range dynamically because some values are out of that min/max-range you always get the initially one after a reset. Is that how it's suppost to be? 
As I understand it, it should reset the zoom to the max and min values of the chart/scale and not some initially wanted one. Maybe if you want to have that, there should be a flag which indicates if the boundaries should be deleted or the initial values are set.
Also a bug is when you change the scale type dynamically, the originalOptions for the scale are vanished and this produces an error.
I think the delete version is the better one.

src/chart.zoom.js
```diff
@@ -317,16 +317,14 @@ var zoomPlugin = {
...
 				}
-
-				scale.options = helpers.configMerge(scale.options, scale.originalOptions);
 			});
 
 			helpers.each(chartInstance.data.datasets, function(dataset, id) {
```
jannisgt: This is could also be deleted (row 330-331), if you have multiple datasets and you faded them out, a resetZoom fades them in again. It is a nice feature, but not always wanted. The user should be able to choose between fading in and stay hidden.

PR #148, opened 2018-05-30
Add options to customize drag-to-zoom area

### LINKED ISSUE #106: Unable to set fillStyle color for drag zoom
Currently the drag zoom highlight color is hard coded to `rgba(225,225,225,0.3)`. This color isn't easily visible and changing background colors isn't a viable option in my case. It would be nice to be able to set this color to any color.

jledentu:
Hi,

This PR adds options to customize drag-to-zoom rectangle area:

```js
{
  zoom: {
    drag: {
      borderColor: 'rgba(225,225,225,0.3)'
      borderWidth: 5,
      backgroundColor: 'rgb(225,225,225)'
    }
  }
}
```

Closes #106

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,13 @@ To configure the zoom and pan plugin, you can simply add new config options to y
 		// Enable drag-to-zoom behavior
 		drag: true,
 
+		// Drag-to-zoom rectangle style can be customized
+		// drag: {
+		// 	 borderColor: 'rgba(225,225,225,0.3)'
+		// 	 borderWidth: 5,
+		// 	 backgroundColor: 'rgb(225,225,225)'
+		// },
+
 		// Zooming directions. Remove the appropriate direction to disable 
 		// Eg. 'y' would only allow zooming in the y direction
 		mode: 'xy',
--- a/src/chart.zoom.js
+++ b/src/chart.zoom.js
@@ -565,10 +565,16 @@ var zoomPlugin = {
 
 			var rectWidth = endX - startX;
 			var rectHeight = endY - startY;
+			var dragOptions = chartInstance.options.zoom.drag;
 
-			ctx.fillStyle = 'rgba(225,225,225,0.3)';
-			ctx.lineWidth = 5;
+			ctx.fillStyle = dragOptions.backgroundColor || 'rgba(225,225,225,0.3)';
 			ctx.fillRect(startX, startY, rectWidth, rectHeight);
+
+			if (dragOptions.borderWidth > 0) {
+				ctx.lineWidth = dragOptions.borderWidth;
+				ctx.strokeStyle = dragOptions.borderColor || 'rgba(225,225,225)';
+				ctx.strokeRect(startX, startY, rectWidth, rectHeight);
+			}
 		}
 
 		ctx.rect(chartArea.left, chartArea.top, chartArea.right - chartArea.left, chartArea.bottom - chartArea.top);


README.md
```diff
@@ -41,6 +41,13 @@ To configure the zoom and pan plugin, you can simply add new config options to y
 		// Enable drag-to-zoom behavior
 		drag: true,
 
+		// Drag-to-zoom rectangle style can be customized
+		// drag: {
```
benmccann: no need to have these lines commented out

jledentu: Fixed

benmccann:
@jledentu sorry for the very long delay in reviewing. Would you be able to rebase this PR?

src/chart.zoom.js
```diff
@@ -527,10 +527,16 @@ var zoomPlugin = {
 			var endX = Math.max(beginPoint.clientX, endPoint.clientX) - offsetX;
 			var rectWidth = endX - startX;
 
+			var dragOptions = chartInstance.options.zoom.drag;
 
-			ctx.fillStyle = 'rgba(225,225,225,0.3)';
-			ctx.lineWidth = 5;
+			ctx.fillStyle = dragOptions.backgroundColor || 'rgba(225,225,225,0.3)';
```
benmccann: This line seems like it could fail if someone sets `drag: true`. It seems potentially problematic that `drag` can contain either a boolean or an object

jledentu: Mmh if `drag: true`, then `dragOptions.backgroundColor` will be `undefined` and the default color will be used, so I don't think it's an issue. Nevertheless, for better readability and maintainability, we could use another property as you suggest. `dragOptions` for instance?

```js
{
  zoom: {
    dragOptions: {
      borderColor: 'rgba(225,225,225,0.3)'
      borderWidth: 5,
      backgroundColor: 'rgb(225,225,225)'
    }
  }
}
```

benmccann: Ah. You're right. The code will work as written. I'm fine with leaving it as is

If we did want to change it another option could be to do something like the below, but I think it's probably fine the way you have it
```
{
  zoom: {
    drag: {
      enabled: true
      borderColor: 'rgba(225,225,225,0.3)'
      borderWidth: 5,
      backgroundColor: 'rgb(225,225,225)'
    }
  }
}
```

src/chart.zoom.js
```diff
@@ -527,10 +527,16 @@ var zoomPlugin = {
...
 			ctx.fillRect(startX, yAxis.top, rectWidth, yAxis.bottom - yAxis.top);
+
+			if (dragOptions.borderWidth > 0) {
```
benmccann: We should keep it working when `borderWidth` is `undefined`. You could change these two lines to:
```
if (dragOptions.borderWidth !== 0) {
	ctx.lineWidth = dragOptions.borderWidth || 5;
```

src/chart.zoom.js
jledentu: Ok so I leave as its is. Checking `drag.enabled` could introduce breaking changes in some configurations (if any user set anything other than a boolean as `drag`), it may be handled in another MR.

src/chart.zoom.js
jledentu: If we do this, we'll introduce a change since there is no border currently. Are you OK with this?

jledentu:
> @jledentu sorry for the very long delay in reviewing. Would you be able to rebase this PR?

Rebase done

src/chart.zoom.js
benmccann: Ah, you're right. I saw we had `ctx.lineWidth = 5` before, but I guess that wasn't used

benmccann:
Thanks again for contributing!

And sorry again for the long delay as well. We're looking for more maintainers. Would you be interested?

jledentu:
@benmccann Sure, I'd be happy to contribute!

PR #150, opened 2018-05-30
Fix mouseup outside chart element

### LINKED ISSUE #149: Drag-to-zoom is not applied when mouseup outside the chart
Hi,

This issue is a **bug** report.

Demo: https://codepen.io/anon/pen/oygLQG

* Press the mouse (_mousedown_) on the chart -> drag-to-zoom begins
* Move the cursor outside the chart
* Release the mouse (_mouseup_) outside the chart -> zoom is not applied, but if you move the mouse on the chart, the drag rectangle is still displayed and updated

That's because the `mouseup` listener is added on the chart element. I think it should be added on the document instead.

jledentu:
Add the `mouseup` listener on the document rather than on the chart element.

Fix #149

### DIFF
--- a/src/chart.zoom.js
+++ b/src/chart.zoom.js
@@ -389,7 +389,7 @@ var zoomPlugin = {
 					chartInstance.zoom._dragZoomEnd = null;
 				}
 			};
-			node.addEventListener('mouseup', chartInstance.zoom._mouseUpHandler);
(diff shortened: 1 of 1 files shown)


JanOonk:
I can confirm this works.

benmccann:
I kind of wonder if it should cancel the zoom instead if released outside the chart area. In any case, this behavior is better than what we have today, so I'll merge

Issue #151, opened 2018-05-31
Fix category scale Zoom/Pan proportion problems


subogero:
Plus panning on indexScale when zoomed in also causes zoom.

Zoom: the `zoom` parameter is a proportion. On linearScale it does work like that, but on the indexScale it's adding/removing 1 label at a time. Totally broken with large datasets. Make it work like linearScale.

Pan: calculate `dIndex` first, limit if `0 .. lastLabelIndex` range would be exceeded, *then* modify `minIndex`, `maxIndex` with the exact same `dIndex`.

Pan on indexScale currently unusable with large datasets. Make pan distance proportional to mouse/touch gesture size. 

Probably fixes some other related issues.

Pull request coming immediately.

subogero:
Also see https://www.reddit.com/r/webdev/comments/8nqcq0/best_js_lib_for_charts/

benmccann:
Also reported in https://github.com/chartjs/chartjs-plugin-zoom/issues/190

benmccann:
Pending PR: https://github.com/chartjs/chartjs-plugin-zoom/pull/152

kurkle:
I believe this was fixed in #476

PR #155, opened 2018-06-22
Fix drag-to-zoom (rebased)

rasmusfaber:
This fixes "Drag-zoom does not zoom to marked area #154"

PR #147, opened 2018-05-30
Enable zoom on vertical axis

jledentu:
Hi,

This PR allows to drag-zoom vertically and make zoom on X axis independent from zoom on Y axis.
The `doZoom` now accepts two parameters to specify horizontal zoom and vertical zoom. 

Some refactoring may be possible to reduce duplicated instructions. Suggestions are welcome. 

Closes #146

PR #161, opened 2018-08-20
Fix the calculation of the zoom center coordinate when dragging

nagix:
### Problem

When the `drag` option is set to `true` and dragging to select a part of the chart area, the zoomed area is not correct. For example, select the x-axis from 0 to 100 in the following chart.
https://jsfiddle.net/nagix/qb3ycp9v/

### Solution
This PR fixes the calculation of the zoom center coordinate when dragging.

* When selecting an area between the left edge and some point in the chart, the zoom center should be the left edge
* When selecting an area between the right edge and some point in the chart, the zoom center should be the right edge


nagix:
#155 only fixes the problem in the time scale, but this solution can apply to any type of the scales and is much simpler.

benmccann:
I'm confused as to what you mean by "zoom center" here. Wouldn't the center be halfway between the left and right edges of the dragged area? Center by definition means the middle, so how can it be the left or right edge?

nagix:
No, the center doesn't mean the middle between the left and right edges, but the focal point of the zoom. The data point at the "zoom center" doesn't change after zooming.

The current implementation regards the middle between the left and right edges of the user selection as the focal point, but that is true only when the middle point is the same as the center of the chart area.

When looking at the first example in the diagram above, the data point at the left edge of the selection should be mapped to the left edge of the chart area, and the that at the right edge of the selection should be mapped to the right edge of the chart area. In this case, the the focal point is at the left edge of the selection.

benmccann:
I don't understand what you mean by focal point or why a focal pointed should be involved

nagix:
> I don't understand what you mean by focal point or why a focal pointed should be involved

That's because the function `doZoom(chartInstance, zoom, center, whichAxes)` accepts the focal point on the axis as `center`.

I'm not sure if the term "focal point" or "zoom center" is appropriate here, but let's look at an actual example. Assume we have a chart with a linear scale in the range between 0 and 100. Then, select the range between 0 and 50 for zooming.

We expect that the new scale will have the range between 0 and 50, but in the current implementation, it will be between 12.5 and 62.5. What happens is that the zoom magnification is calculated correctly at https://github.com/chartjs/chartjs-plugin-zoom/blob/a8de40507e9546d18abe9454fbacbe67aadf5fe5/src/chart.zoom.js#L386 and `zoom` is set to 1.5 (the actual zoom magnification will be the reciprocal of `zoom - 1`), but `center` is set to `25 + startX` at https://github.com/chartjs/chartjs-plugin-zoom/blob/a8de40507e9546d18abe9454fbacbe67aadf5fe5/src/chart.zoom.js#L393-L396 which is the center of the selection, not the zoom focal point. As a result, the position on the scale for the value 25 won't change but the distances from `center` will be magnified by 2, then `ticks.min` is set to 12.5 and `ticks.max` to 62.5. https://github.com/chartjs/chartjs-plugin-zoom/blob/a8de40507e9546d18abe9454fbacbe67aadf5fe5/src/chart.zoom.js#L144-L152

What we want is to map the left edge of the selection to the left edge of the new scale and the right edge of the selection to the right edge of the new scale. When we select the area between 0 and 50, the left edge (= 0) doesn't change, so 0 should be passed as `center`. As a result, `ticks.min` is set to 0 and `ticks.max` to 50 as expected.

To generalise this, the calculation of `center` would be like this https://github.com/chartjs/chartjs-plugin-zoom/blob/97d11f5129dbc806089e99ec5f3f7785c7a4aa67/src/chart.zoom.js#L382-L385

dmrickey:
This worked perfectly for me.  On my time chart, when I clicked and dragged it was not zooming in to the proper selection.  I pulled this down and registered the code directly instead of the published package and it now zooms in to the proper selection.

CNA-Bld:
IMHO this correctly fixes the problem. The "focal point" (`center`) mentioned above should probably be called "fixed point" though.

When using a mouse wheel, we get it from the mouse pointer coordinates so the zooming happens "around" it and it fells natural. When dragging, however, we have to do the calculation explained above.

simonbrunel:
> I'm not sure if the term "focal point" or "zoom center" is appropriate here

I saw it called ["anchor"](http://doc.qt.io/qt-5/qgraphicsview.html#ViewportAnchor-enum) or ["origin"](http://doc.qt.io/qt-5/qml-qtquick-scale.html#origin-prop), which is the point that remains unchanged during transformations.

Anyway, I'm not sure this PR is waiting for a name :) The logic makes sense but I don't know the code of this plugin to approve this PR. Since @dmrickey and @CNA-Bld confirmed that it works as expected, feel free to merge.

benmccann:
Sorry, I've been meaning to come back to this PR to review and get it merged. I'll take another pass now

benmccann:
I just pulled down this PR and built it locally. I'm not sure it works exactly as I'd expect in all cases

In samples/zoom-bar-x.html, my mouse is not remaining the focal point. First zoom all the way out on this sample first to show more data on the chart. Now let go and start a new zoom inward. I'd expect the date that my mouse is hovering over to remain under my mouse. Instead it starts moving sideways

Also, in samples/zoom-log.html put your mouse right next to a point and zoom in. I'd expect to zoom in on that point, but that's not what happens. The chart moves away from my mouse both horizontally and vertically.

benmccann:
Sorry, I just realized this PR only affects when dragging and those bugs are unrelated. Let me do some more testing

benmccann:
I think this may only be working in the x-axis direction. I edited samples/zoom.html to set `pan.enabled: false` and `zoom.drag: true`.

I then clicked and dragged from (0, 0) to (20, 20). The zoom looked correct in the x-axis. However, the y-axis zoomed to (-100, 80). I removed `zoom.limits` in case it was interfering, but the behavior didn't change at all.

nagix:
@benmccann The lines below indicate dragging is only supported for horizontal axis: https://github.com/chartjs/chartjs-plugin-zoom/blob/97d11f5129dbc806089e99ec5f3f7785c7a4aa67/src/chart.zoom.js#L348-L350

We may be able to support dragging for vertical axis, but I would like to focus on fixing the issue in the current implementation in this PR. I will add notes about this limitation to README.

nagix:
Actually, #147 is trying to support drag-zoom on vertical axis.

benmccann:
Thanks for pointing out #147. I went ahead and merged that one to address it. Would you be able to rebase this PR?

benmccann:
This PR might also fix https://github.com/chartjs/chartjs-plugin-zoom/issues/151?

nagix:
@benmccann Thanks for merging #147 first. That makes this PR simpler, and rebase was done.

This PR only handles dragging, so #151 seems to be a different issue.

benmccann:
Thanks for this @nagix !

PR #178, opened 2018-11-20
Added onPan/onZoom args

Techn1x:
Knowing how the chart was zoomed or panned will probably be useful information (at least, it was for me).

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ To configure the zoom and pan plugin, you can simply add new config options to y
 		},
 		// Function called once panning is completed
 		// Useful for dynamic data loading
-		onPan: function({chart, deltaX, deltaY}) { console.log(`I was panned!!! deltaX: ${deltaX}, deltaY: ${deltaY}`); }
+		onPan: function({chart}) { console.log(`I was panned!!!`); }
 	},
 	
 	// Container for zoom options
@@ -66,7 +66,7 @@ To configure the zoom and pan plugin, you can simply add new config options to y
 		},
 		// Function called once zooming is completed
 		// Useful for dynamic data loading
-		onZoom: function({chart, zoomX, zoomY, center, whichAxes}) { console.log(`I was zoomed!!! zoomX: ${zoomX} zoomY: ${zoomY}`); }
+		onZoom: function({chart}) { console.log(`I was zoomed!!!`); }
 	}
 }
 ```
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ To configure the zoom and pan plugin, you can simply add new config options to y
 		},
 		// Function called once panning is completed
 		// Useful for dynamic data loading
-		onPan: function() { console.log('I was panned!!!'); }
+		onPan: function({chart, deltaX, deltaY}) { console.log(`I was panned!!! deltaX: ${deltaX}, deltaY: ${deltaY}`); }
 	},
 	
 	// Container for zoom options
@@ -66,7 +66,7 @@ To configure the zoom and pan plugin, you can simply add new config options to y
 		},
 		// Function called once zooming is completed
 		// Useful for dynamic data loading
-		onZoom: function() { console.log('I was zoomed!!!'); }
+		onZoom: function({chart, zoomX, zoomY, center, whichAxes}) { console.log(`I was zoomed!!! zoomX: ${zoomX} zoomY: ${zoomY}`); }
 	}
 }
 ```
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -200,7 +200,7 @@ function doZoom(chart, percentZoomX, percentZoomY, focalPoint, whichAxes) {
 		chart.update(0);
 
 		if (typeof zoomOptions.onZoom === 'function') {
-			zoomOptions.onZoom();
+			zoomOptions.onZoom({chart: chart});
 		}
 	}
 }
@@ -278,7 +278,7 @@ function doPan(chartInstance, deltaX, deltaY) {
 		chartInstance.update(0);
 
 		if (typeof panOptions.onPan === 'function') {
-			panOptions.onPan();
+			panOptions.onPan({chart: chartInstance});
 		}
 	}
 }


benmccann:
What do you think about putting these arguments in an object that's passed to the function instead? That'd be similar to the context object for options that take functions in the main library: https://www.chartjs.org/docs/latest/general/options.html#option-context. It also would allow you to add more easily without the parameters list becoming too long and make it easier for callers who care about only one of the arguments

Techn1x:
Sounds good to me. I'll implement that when I get some free time

benmccann:
Cool. Let me know and then I'll merge it. I like the idea of passing some info to the callback

Techn1x:
Rebased and args are now passed as an object.

I wasn't able to test it though because gulp seems to be broken now? I get an error about an when I run `gulp` like this `throw er; // Unhandled 'error' event`. But that is most likely unrelated to my changes.

benmccann:
I think it's failing because the code isn't valid syntax. E.g. `{chartInstance, deltaX, deltaY}` should be:

```
{
  chartInstance: chartInstance,
  deltaX: deltaX,
  deltaY: deltaY
}
```

Techn1x:
Seems you were right. It's valid syntax as of ES2015 when the key matches the variable name... but I guess lint doesn't like it

I've fixed the syntax

src/chart.zoom.js
```diff
@@ -198,7 +198,13 @@ function doZoom(chartInstance, zoomX, zoomY, center, whichAxes) {
 		chartInstance.update(0);
 
 		if (typeof zoomOptions.onZoom === 'function') {
-			zoomOptions.onZoom();
+			zoomOptions.onZoom({
+				chartInstance: chartInstance,
```
benmccann: let's call this `chart` on the lefthand site. will match what we do in the main repo and is simpler

benmccann: same below for the pan callback

src/chart.zoom.js
```diff
@@ -198,7 +198,13 @@ function doZoom(chartInstance, zoomX, zoomY, center, whichAxes) {
...
+				zoomX: zoomX,
```
benmccann: What are `zoomX` and `zoomY`?  (I took over maintaining this repo, so don't know all the code). Can we remove them and get by with just `center`? If not, can we rename or document to clarify?

src/chart.zoom.js
```diff
@@ -198,7 +198,13 @@ function doZoom(chartInstance, zoomX, zoomY, center, whichAxes) {
...
+				zoomY: zoomY,
+				center: center,
+				_whichAxes: _whichAxes
```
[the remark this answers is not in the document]
benmccann: Actually, do we even need to expose this? Won't the user know what it is from the zoom mode option?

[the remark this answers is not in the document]
Techn1x: _whichAxes is not necessarily equal to the zoom mode. Eg zoom mode can be 'xy' but the zoom performed might be just 'x', I think

src/chart.zoom.js
```diff
@@ -198,7 +198,13 @@ function doZoom(chartInstance, zoomX, zoomY, center, whichAxes) {
 		chartInstance.update(0);
 
 		if (typeof zoomOptions.onZoom === 'function') {
-			zoomOptions.onZoom();
+			zoomOptions.onZoom({
+				chart: chartInstance,
+				zoomX: zoomX,
+				zoomY: zoomY,
+				center: center,
```
benmccann: I'm not sure `center` is a good name for this. It's probably better called `focalPoint`. See the discussion here https://github.com/chartjs/chartjs-plugin-zoom/pull/161#issuecomment-417947342. CC @nagix

src/chart.zoom.js
benmccann: I looked into what this is doing and am proposing to rename to `percentZoomX` and `percentZoomY` in https://github.com/chartjs/chartjs-plugin-zoom/pull/192

benmccann:
Thinking about this more, I wonder if it's really needed. Can you check the scale `min` and `max` after zoom / pan instead? What's your use case? 

I'm a bit hesitant about the zoom arguments. I think we may end up refactoring some of this code so that the current arguments are no longer relevant in all cases. What I'm thinking is that if you do a drag zoom then right now we convert that to a zoom percentage and focal point and then find the area zoom based off of that, which seems complicated compared to just zooming to the dragged area. I'd like those arguments to go away in that case.

src/chart.zoom.js
benmccann: I've proposed to rename the argument in https://github.com/chartjs/chartjs-plugin-zoom/pull/192

jledentu:
@benmccann I agree. Returning `min`/`max` of each transformed scale seems more useful, especially in drag mode (you can compute the zoom factor anyway).

Example of use case: adapt the unit depending on the axis bounds (https://github.com/chartjs/chartjs-plugin-zoom/issues/179). Here `min`/`max` are relevant values.

Techn1x:
I'll admit, in my use case of the doZoom() I'm not actually using the arguments, just getting the new min/max from the chart instance and using that. So I don't know a whole lot about the zoom args.

Let me know what you think the args should be when your mind is made up and I'll change it to that then rebase/squash

benmccann:
I would say that we don't pass any argument to `doZoom` since you can read the scale `min` and `max` already today

Even for the pan case, is there a reason you don't just read the scale `min` and `max`? I'm wondering if we need this PR at all actually

Techn1x:
That might be the case but I think passing at least the chart instance is helpful as an arg to `onZoom` and `onPan`

benmccann:
Ok. You're right that passing the chart instance would probably be helpful. In the case that the user has multiple charts on the page they'll know which one was acted upon

benmccann:
@Techn1x can you rebase this PR and then pass only `chart` into the callback? Thanks!

Techn1x:
@benmccann Rebased, squashed, and added `chart` as the only arg for `onPan` and `onZoom`

benmccann:
Thanks @Techn1x !

rahul1500:
@benmccann these changes are not released currently

benmccann:
They will be included in the next release (hopefully in the next couple of weeks)

PR #192, opened 2019-01-24
Add jsdoc for doZoom

benmccann:
@Techn1x is proposing to expose some of these parameters to users in https://github.com/chartjs/chartjs-plugin-zoom/pull/178, so I wanted to make sure they were well understood and documented. I think some of the parameter names were unclear before

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -153,21 +153,28 @@ function zoomScale(scale, zoom, center, zoomOptions) {
 	}
 }
 
-function doZoom(chartInstance, zoomX, zoomY, center, whichAxes) {
-	var ca = chartInstance.chartArea;
-	if (!center) {
-		center = {
+/**
+ * @param chart The chart instance
+ * @param {number} percentZoomX The zoom percentage in the x direction
+ * @param {number} percentZoomY The zoom percentage in the y direction
+ * @param {Object} focalPoint The x and y coordinates of zoom focal point. The point which doesn't change while zooming. E.g. the location of the mouse cursor when "drag: false"
+ * @param {string} whichAxes `xy`, 'x', or 'y'
+ */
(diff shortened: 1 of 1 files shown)


src/chart.zoom.js
```diff
@@ -158,21 +158,28 @@ function zoomScale(scale, zoom, center, zoomOptions) {
...
 
-function doZoom(chartInstance, zoomX, zoomY, center, whichAxes) {
-	var ca = chartInstance.chartArea;
-	if (!center) {
-		center = {
+/**
+ * @param chart The chart instance
+ * @param {number} percentZoomX The zoom percentage in the x direction
+ * @param {number} percentZoomY The zoom percentage in the y direction
+ * @param {Object} focalPoint The x and y coordinates of zoom focal point. The point which doesn't change while zooming. E.g. the location of the mouse cursor when "drag: false"
```
jledentu: I prefer the word `center` which I think more common.

jledentu: You could document the type of `focalPoint` as below:

```
@param {{x: number, y: number}} focalPoint
```

benmccann: Good idea. Done

benmccann: `center` is a more common term, but I also think it's a bit misleading. `center` means the point in the middle, but this point is often not in the middle. Using a less common term may actually be useful because it'll make people stop and consider what this point is, which I think is valuable because it's not immediately clear especially in the drag to zoom case

@nagix and I had some discussion about this here where he first used the term focal point: https://github.com/chartjs/chartjs-plugin-zoom/pull/161#issuecomment-417947342

jledentu: Ok, I get your point.

PR #193, opened 2019-01-24
Use Rollup to build the library

jledentu:
As a first iteration to integrate into this plugin the tools used in the core and other plugins, this PR adds [Rollup](https://github.com/rollup/rollup) as module bundler to build the library. It allows to use ES modules, so we'll be able to split the plugin into several files.

- Rewrite the `build` task using Rollup and some Rollup plugins
- Remove old libraries used for building
- The dist files are now generated into `/dist` and unversioned 
- The old dist files have been removed
- Rename the source file to `plugin.js` (as in `chartjs-plugin-datalabels`)

Notes:

- This PR contains a breaking change since the distribution files are moved to `/dist`, so it should imply a `0.7` version
- I think the Travis config should be updated to generate the dist files automatically on release (in a separate PR?)


rollup.config.js
```diff
@@ -0,0 +1,55 @@
...
+	{
+		input: 'src/plugin.js',
+		output: {
+			name: 'ChartZoom',
+			file: `dist/${pkg.name}.js`,
+			banner: banner,
+			format: 'umd',
+			indent: false,
+			globals: {
+        'chart.js': 'Chart',
```
benmccann: should use tabs here instead of spaces

rollup.config.js
```diff
@@ -0,0 +1,55 @@
...
+		},
+		plugins: [
+			terser({output: {comments: 'some'}})
+		],
+		external: [
+			'chart.js',
+			'hammerjs'
+		]
+	}
+];
```
benmccann: can you add a new line at the end of the file?

package.json
```diff
@@ -14,17 +14,14 @@
...
-    "gulp-replace": "^1.0.0",
-    "gulp-streamify": "^1.0.2",
-    "gulp-uglify": "^1.5.3",
     "gulp-util": "^3.0.8",
     "inquirer": "^6.2.0",
-    "merge-stream": "^1.0.1",
-    "semver": "^5.5.1",
-    "vinyl-source-stream": "^2.0.0"
+    "rollup": "^1.1.2",
+    "rollup-plugin-commonjs": "^9.2.0",
```
benmccann: I don't think this is needed

gulpfile.js
```diff
@@ -1,57 +1,33 @@
...
-    .pipe(streamify(uglify({
-      preserveComments: 'some'
-    })))
-    .pipe(streamify(concat('chartjs-plugin-zoom.min.js')))
-    .pipe(gulp.dest(outDir));
-
-  return nonBundled;
-
+function buildTask(done) {
+  run('rollup/bin/rollup', ['-c'], done);
```
benmccann: tabs vs spaces

package.json
jledentu: You're right, it was before I replaced `require` by `import`.

package.json
```diff
@@ -14,17 +14,14 @@
...
+    "rollup-plugin-node-resolve": "^4.0.0",
```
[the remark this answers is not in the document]
jledentu: Fixed

rollup.config.js
jledentu: Fixed

gulpfile.js
jledentu: Argh, I fixed this and added a `.editorconfig` file (as in the Chart.js core).

package.json
jledentu: You're right, it was before I replaced `require` by `import`.

package.json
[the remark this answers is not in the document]
jledentu: You're right, it was before I replaced `require` by `import`.

rollup.config.js
```diff
@@ -0,0 +1,55 @@
...
+		input: 'src/plugin.js',
+		output: {
+			name: 'ChartZoom',
+			file: `dist/${pkg.name}.js`,
+			banner: banner,
+			format: 'umd',
+			indent: false,
+			globals: {
+        'chart.js': 'Chart',
+        'hammerjs': 'Hammer'
```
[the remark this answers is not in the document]
jledentu: I think that we can include `hammerjs` in the generated bundle for simpler usage. People will still be able to generate a lighter bundle (if they want to) by importing directly the source file.

package.json
```diff
@@ -14,17 +14,12 @@
...
-    "gulp-insert": "^0.5.0",
-    "gulp-replace": "^1.0.0",
-    "gulp-streamify": "^1.0.2",
-    "gulp-uglify": "^1.5.3",
     "gulp-util": "^3.0.8",
     "inquirer": "^6.2.0",
-    "merge-stream": "^1.0.1",
-    "semver": "^5.5.1",
-    "vinyl-source-stream": "^2.0.0"
+    "rollup": "^1.1.2",
```
benmccann: I believe you can remove the `browserify` dependency now

rollup.config.js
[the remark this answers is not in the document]
jledentu: @benmccann I included `hammerjs`as internal dep. As expected, it increases the weight of the bundle.

package.json
jledentu: Right! Done.

benmccann:
I'm not sure I fully comprehend how moving the location of the compiled files would be a breaking change.

If you include via npm, you should not have a problem because the `main` attribute in package.json tells you where the entry point is. And if you include via `<script>` tag then you are manually pointing at the compiled file.

simonbrunel:
> If you include via npm, you should not have a problem because the main attribute in package.json

Previous versions have `main: "src/chart.zoom.js"` as entry point, which isn't the build file that can run in the browser, so there is a high probability some users included `./chartjs-plugin-zoom.min.js` instead of relying on the `package.json:main`. In this case, this change will break their projects.

> If you include via <script> tag then you are manually pointing at the compiled file

That will break because the compiled file will not be available anymore at this location. Note that some CDNs allow to access files using semver range, for example: 

https://cdn.jsdelivr.net/npm/chartjs-plugin-zoom/chartjs-plugin-zoom.min.js  `// latest major`
https://cdn.jsdelivr.net/npm/chartjs-plugin-zoom@0/chartjs-plugin-zoom.min.js `// latest minor`

While the first form (`latest major`) should not be used, the second one (`latest minor`) is pretty common since it's similar to `^0.6.0`. Both will break starting 0.7.0.

So I'm just saying that considering the age of this plugin, we should try to prevent these situations.

jledentu:
SemVer allows breaking changes in 0.x versions, considered as major versions (`^0.6.0` won't install `0.7.0`). But I admit that I didn't consider the CDN users, and we have to not break the app of these users.

Do we consider the API of this plugin is mature enough to publish a 1.0.0 version? I'm not sure, and I think we should fix some issues before. So @simonbrunel I like your suggestion to keep a copy of the dist files in the root folder. Do you agree @benmccann?

@simonbrunel Thanks, I'll make another PR to update the Travis config in a couple of days.

benmccann:
Yeah, let's explore the idea of copying files to the root directory. I agree that it'd make sense to fix some more issues before declaring 1.0

Techn1x:
FWIW - in my project I was pointing to my PR for pan/zoom args (which has now been rebased to include this PR) and I have run into some of the issues mentioned

I'm not super across JS imports and such but I was relying on the `/dist` folder for `chartjs-plugin-zoom.min.js`, but now this file no longer exists. I tried importing the `plugin.js` but that didn't work either. When I do an `npm install` on my project, it doesn't seem to build the `/dist` folder for the plugin. So basically I don't know what to import... any help? Or will the `/dist` files be back for the next release? If thats the case I can build them manually until that happens

Techn1x:
That seems strange that I need to add an extra build tool to my project to accommodate a chartjs plugin...
I tried to run `gulp build` in my project but I think I need a `gulpfile.js`? What do I need to put in there?

benmccann:
You don't use `gulp` to build your project, but rather to build this library. It will be built each time it is released, but if you'd like to use it between releases then you will need to build it yourself. This is the same way the main https://github.com/chartjs/Chart.js repo works

Techn1x:
Ah yep that's what I thought originally but I didn't convey that properly. 

I'll build this plugin myself with gulp as I have been when using a branch or PR, but once the release is cut I'll be able to use the release build, thanks

benmccann:
@jledentu do you think you'll be able to send that PR to copy files to the root directory?

jledentu:
@benmccann Yes! Sorry, I didn't have time to finalize the PR, but I'm working on it and I expect to submit it in a few days.

The PR will copy the files to the root directory, but will also allow to generate the release zip + the bower conf, as in the root project.

benmccann:
That's great, thanks!

jledentu:
@benmccann @simonbrunel 

> It must be updated to include built files in tags and npm releases. You will find everything needed in the datalabels plugin and/or the core repo. I will configure Travis environment variables once your PR is ready.

This other PR (#202) brings the changes you suggested.

PR #194, opened 2019-01-24
lint samples

benmccann:


benmccann:
@jledentu would you be able to approve this PR for me? I can't merge any code until it has at least 1 approval. Thanks!

.htmllintrc
```diff
@@ -0,0 +1,19 @@
+{
+	"indent-style": "tabs",
+	"line-end-style": false,
+	"attr-quote-style": "double",
```
jledentu: Is this needed, since `double` is the default value of `attr-quote-style`?

benmccann: removed. good catch. I just copied the `.htmllintrc` file from one of the other chart.js repos. I guess we could remove it from the others as well

benmccann:
Thanks @jledentu!

PR #196, opened 2019-01-30
Allow switching between wheel/drag modes of the zoom

### LINKED ISSUE #176: toggle options.zoom.drag true <=> false
Would be nice if user could change between drag mode and zoom/pan mode.
Only changing options.zoom.drag value doesn't work.

jledentu:
Closes #176 

This PR allows switching between wheel/drag modes of the zoom (if `options.zoom.drag` is changed, it will be taken into account in next events).

- All event listeners are set up in the `beforeInit` method, and the `options.zoom.drag` setting is checked in the body of these listeners rather than at the initialization.
- I added a `Enable/disable drag mode` in the `zoom-time` sample to test this improvement.
- I also fixed the `Reset Zoom` button in the `zoom-time` sample.

### DIFF
--- a/samples/zoom-time.html
+++ b/samples/zoom-time.html
@@ -45,16 +45,10 @@
 			return window.moment().add(days, 'd').format(timeFormat);
 		}
 
-		/**
-		 * Reset the zoom.
-		 */
 		function resetZoom() {
 			window.myLine.resetZoom();
 		}
 
-		/**
-		 * Toggle the `drag` mode of the zoom.
-		 */
 		function toggleDragMode() {
 			var zoomOptions = window.myLine.options.zoom;
 			zoomOptions.drag = !zoomOptions.drag;
--- a/samples/zoom-time.html
+++ b/samples/zoom-time.html
@@ -19,6 +19,7 @@
 <body>
 	<div style="width:75%;">
 		<button onclick="resetZoom()">Reset Zoom</button>
+		<button id="drag-switch" onclick="toggleDragMode()">Disable drag mode</button>
 		<canvas id="canvas"></canvas>
 	</div>
 	<script>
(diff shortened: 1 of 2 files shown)


samples/zoom-time.html
```diff
@@ -19,6 +19,7 @@
 <body>
 	<div style="width:75%;">
 		<button onclick="resetZoom()">Reset Zoom</button>
+		<button id="drag-switch" onclick="toggleDragMode()">Disable drag mode</button>
```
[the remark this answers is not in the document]
benmccann: Ah. Nevermind. I hadn't read down to where you're toggling the text :-)

src/plugin.js
```diff
@@ -364,66 +364,69 @@ var zoomPlugin = {
...
+		chartInstance.zoom._mouseMoveHandler = function(event) {
+			if (chartInstance.options.zoom.drag && chartInstance.zoom._dragZoomStart) {
+				chartInstance.zoom._dragZoomEnd = event;
+				chartInstance.update(0);
+			}
+		};
+		node.addEventListener('mousemove', chartInstance.zoom._mouseMoveHandler);
+
+		chartInstance.zoom._mouseUpHandler = function(event) {
+			if (chartInstance.options.zoom.drag && chartInstance.zoom._dragZoomStart) {
```
benmccann: I'd probably prefer to do the inverse to save from having to indent the rest of the code as much:

```
if (!chartInstance.options.zoom.drag || !chartInstance.zoom._dragZoomStart) {
  return;
}
```

benmccann: But probably only for a longer method like this one. For a method that's only a few lines, what you have now is shorter

src/plugin.js
```diff
@@ -364,66 +364,69 @@ var zoomPlugin = {
...
-				if (chartInstance.zoom._dragZoomStart) {
-					chartInstance.zoom._dragZoomEnd = event;
-					chartInstance.update(0);
+		chartInstance.zoom._mouseMoveHandler = function(event) {
+			if (chartInstance.options.zoom.drag && chartInstance.zoom._dragZoomStart) {
+				chartInstance.zoom._dragZoomEnd = event;
+				chartInstance.update(0);
+			}
+		};
+		node.addEventListener('mousemove', chartInstance.zoom._mouseMoveHandler);
```
benmccann: I forgot we have a `mousemove` handler. `mousemove` can fire hundreds of times a second constantly, so we should be a lot more careful than with `mouseup`, `mousedown`, etc. I'd be more comfortable for performance reasons if went the route of switching the listeners on and off instead eventhough it would be a bit more complicated.

jledentu: Good point! I fixed and now add the `mousemove` listener on `mousedown`. Other listeners stay initialized as it is.

Issue #153, opened 2018-06-07
Zooming out on logarithmic axis clears the chart


RaitzeR:
If you try to zoom out on a logarithmic axis, the chart data will disappear. I think this is because it tries to get a negative value for the scale.
There is no problems with panning though. It correctly gets the new scale.

  ```
options: {
                scales: {
                    xAxes: [{
                        type: 'linear',
                        ticks: {
                            userCallback: function(tick) {
                                return timeConverter(tick, false);
                            },
                            autoSkip: false,
                            maxRotation: 20,
                            minRotation: 20
                        },
                        position: 'bottom'
                    }],
                    yAxes: [{
                        type: 'logarithmic',
                        ticks: {
                            userCallback: function(tick) {
                                var remain = tick / (Math.pow(10, Math.floor(Chart.helpers.log10(tick))));
                                if (remain === 1) {
                                    return sanitizeNumber(tick).toString();
                                }
                                return '';
                            },
                        },
                    }]
                },
                tooltips: {
                    callbacks: {
                        label: function(tooltipItem, chartData) {
                            return "Time: "+ timeConverter(tooltipItem.xLabel, true) +" - Price:"+ sanitizeNumber(tooltipItem.yLabel) +""
                        }
                    }
                },
                pan: {
                    enabled: true,
                    mode: 'xy',
                },
                zoom: {
                    enabled: true,
                    mode: 'xy',
                }
```

**Comment by BDominik** [comment]
I have exactly the same problem.
Did you solved this?
I'm using logarithmic scaling on the x axis only.

**Comment by BDominik** [comment]
Actually if you set the zoom and pan ranges, than it's working
`rangeMin: {
    x: 10,
    y: 0.0193
},
rangeMax: {
    x: 8640,
    y: 1.4
}`

**Comment by jannisgt** [comment]
Hi,

I found a solution for that problem. This is the actual function for numerical scale zoom.
```
function zoomNumericalScale(scale, zoom, center, zoomOptions) {
	var range = scale.max - scale.min;
	var newDiff = range * (zoom - 1);

	var cursorPixel = scale.isHorizontal() ? center.x : center.y;
	var minPercent = (scale.getValueForPixel(cursorPixel) - scale.min) / range;
	var maxPercent = 1 - minPercent;

	var minDelta = newDiff * minPercent;
	var maxDelta = newDiff * maxPercent;

[...]

PR #199, opened 2019-02-06
Fix zooming clears logarithmic scale

jannisgt:
Existing error: Logarithmic scale gets cleared out by zooming out.

Error diagnosis: Minimum value of scale was getting a negative value, logarithmic scales are not able to show negative values -> gets cleared out.

Bug fixing: When the scale would get a negative value it will be substituted by 0.

Closes https://github.com/chartjs/chartjs-plugin-zoom/issues/153

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -132,6 +132,7 @@ function zoomTimeScale(scale, zoom, center, zoomOptions) {
 }
 
 function zoomNumericalScale(scale, zoom, center, zoomOptions) {
+	var tickOpts = scale.options.ticks;
 	var range = scale.max - scale.min;
 	var newDiff = range * (zoom - 1);
 
@@ -141,6 +142,10 @@ function zoomNumericalScale(scale, zoom, center, zoomOptions) {
 
 	var minDelta = newDiff * minPercent;
 	var maxDelta = newDiff * maxPercent;
+
+	if (scale.type === 'logarithmic') {
+		tickOpts.min = tickOpts.min < 0 ? 0 : tickOpts.min;
+	}
 
 	scale.options.ticks.min = rangeMinLimiter(zoomOptions, scale.min + minDelta);
 	scale.options.ticks.max = rangeMaxLimiter(zoomOptions, scale.max - maxDelta);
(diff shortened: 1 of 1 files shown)


src/plugin.js
```diff
@@ -142,6 +143,10 @@ function zoomNumericalScale(scale, zoom, center, zoomOptions) {
 	var minDelta = newDiff * minPercent;
 	var maxDelta = newDiff * maxPercent;
 
+	if (scale.type === 'logarithmic') {
+		tickOpts.min = tickOpts.min < 0 ? 0 : tickOpts.min;
+	}
+
 	scale.options.ticks.min = rangeMinLimiter(zoomOptions, scale.min + minDelta);
```
jledentu: Hi, thanks for your Pull Request!

I tested it (in the sample `zoom-log.html`) but I can't get it work: the bug is still here. I think that the `tickOpts.min =` assignment is overridden by the next `scale.options.ticks.min =` assignment.

You could write something like:

```js
var newMin = rangeMinLimiter(zoomOptions, scale.min + minDelta);

if (scale.type === 'logarithmic') {
  newMin = Math.max(newMin, 0);
}

scale.options.ticks.min = newMin;
scale.options.ticks.max = rangeMaxLimiter(zoomOptions, scale.max - maxDelta);
```

jannisgt: OMG... I forgot the important stuff =(

sorry... now it should be ok

jledentu: No problem. It's actually better. :wink:

benmccann:
I created https://github.com/chartjs/Chart.js/pull/6058 as an alternate fix which would potentially allow us to avoid adding special case logic here. It makes the chart ignore invalid values. However, I'm not sure how it will interact with the zoom plugin as I haven't tested that yet. We may still need this fix if it results in a poor interaction

Would either of you guys be able to test the zoom plugin against https://github.com/chartjs/Chart.js/pull/6058?

jledentu:
@benmccann Nice! Handling this case in the core is actually better!

benmccann:
I'm going to close this since https://github.com/chartjs/Chart.js/pull/6058 was merged

Thanks for raising this issue and proposing the fix

PR #201, opened 2019-02-08
Store original scale options to resetZoom properly

### LINKED ISSUE #197: Adding scales leads to errors in the resetZoom function.
Hi,

when you add dynamically new scales to your chart and uses resetZoom it throws an error. Because there are no originalOptions for scales which were not existing at the point of initialization.

	

```
afterInit: function(chartInstance) {
	helpers.each(chartInstance.scales, function(scale) {
		scale.originalOptions = helpers.clone(scale.options);
	});

	chartInstance.resetZoom = function() {
		helpers.each(chartInstance.scales, function(scale, id) {
			var timeOptions = scale.options.time;
			var tickOptions = scale.options.ticks;

         		 if (timeOptions) {
            			timeOptions.min = scale.originalOptions.time.min;
           			timeOptions.max = scale.originalOptions.time.max;
          		}

          		if (tickOptions) {
            			tickOptions.min = scale.originalOptions.ticks.min;
            			tickOptions.max = scale.originalOptions.ticks.max;
          		}
		});

		helpers.each(chartInstance.data.datasets, function(dataset, id) {
			dataset._meta = null;
		});

		chartInstance.update();
	};

},
```

A remedy would be to add a query if the originalOptions are existing otherwise just delete those options. Like that I think everyone would win.

```
afterInit: function(chartInstance) {
		helpers.each(chartInstance.scales, function(scale) {
			scale.originalOptions = helpers.clone(scale.options);
		});

		chartInstance.resetZoom = function() {
			helpers.each(chartInstance.scales, function(scale, id) {
				var timeOptions = scale.options.time;
				var tickOptions = scale.options.ticks;

                                if (scale.originalOptions) {

					if (timeOptions) {
						timeOptions.min = scale.originalOptions.time.min;
						timeOptions.max = scale.originalOptions.time.max;
					}

					if (tickOptions) {
						tickOptions.min = scale.originalOptions.ticks.min;
						tickOptions.max = scale.originalOptions.ticks.max;
					}
				} else {
		
					if (timeOptions) {
						delete timeOptions.min;
						delete timeOptions.max;
					}

					if (tickOptions) {
						delete tickOptions.min;
						delete tickOptions.max;
					}
				});

				helpers.each(chartInstance.data.datasets, function(dataset, id) {
					dataset._meta = null;
				});

			chartInstance.update();
		};

},
```

jannisgt:
Fixes #197 

With introducing a new data structure `originalOptions`, the scale options can be stored and recalled to reset the zooming/panning.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -342,7 +342,6 @@ var zoomPlugin = {
 	id: 'zoom',
 
 	afterInit: function(chartInstance) {
-		storeOriginalOptions(chartInstance);
 
 		chartInstance.resetZoom = function() {
 			storeOriginalOptions(chartInstance);
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -29,8 +29,17 @@ var defaultOptions = zoomNS.defaults = {
 // Stores the original options of the scales
 var originalOptions = {};
 
-function storeOriginalOptions(scale) {
-	originalOptions[scale.id] = helpers.clone(scale.options);
+function storeOriginalOptions(chart) {
+	helpers.each(chart.scales, function(scale) {
+		if (!originalOptions[scale.id]) {
+			originalOptions[scale.id] = helpers.clone(scale.options);
+		}
+	});
+	helpers.each(originalOptions, function(opt, key) {
+		if (!chart.scales[key]) {
+			delete originalOptions[key];
+		}
+	});
 }
 
 function directionEnabled(mode, dir) {
@@ -178,6 +187,7 @@ function doZoom(chart, percentZoomX, percentZoomY, focalPoint, whichAxes) {
 	var zoomOptions = chart.options.zoom;
 
 	if (zoomOptions && helpers.getValueOrDefault(zoomOptions.enabled, defaultOptions.zoom.enabled)) {
+		storeOriginalOptions(chart);
 		// Do the zoom here
 		var zoomMode = helpers.getValueOrDefault(chart.options.zoom.mode, defaultOptions.zoom.mode);
 		zoomOptions.sensitivity = helpers.getValueOrDefault(chart.options.zoom.sensitivity, defaultOptions.zoom.sensitivity);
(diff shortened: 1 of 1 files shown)


benmccann:
This could leave orphaned scale properties around if you remove a scale. The version I posted in https://github.com/chartjs/chartjs-plugin-zoom/issues/197#issuecomment-461587818 is probably a bit safer

Also, you should delete the `originalOptions` map in the `destroy` method

src/plugin.js
```diff
@@ -26,6 +26,12 @@ var defaultOptions = zoomNS.defaults = {
 		sensitivity: 3
 	}
 };
+// Stores the original options of the scales
```
benmccann: This comment doesn't add much that's not covered by the variable name already. I'd remove it

jannisgt:
With your approach your always overwriting `originialOptions`. And when you zoom/pan then you are going to change min/max  and also saving the new min/max to `originalOptions` I do not think that is the way.

benmccann:
Maybe you could make it something like the code below and move it outside the loop when you call it:
```
function storeOriginalOptions(chart) {
	helpers.each(chart.scales, function(scale) {
		if (!originalOptions[scale.id]) {
			originalOptions[scale.id] = helpers.clone(scale.options)
		}
	}
	helpers.each(originalOptions, function(opt, key) {
		if (!scales[key]) {
			delete originalOptions[key];
		}
	}
}
```

benmccann:
@jannisgt just a reminder about this PR. We will probably release the next version before too long, so I wanted to give you a chance to get this fix in before that. Thanks!

benmccann:
There's one case that I think is not covered. If you add a new scale and then `resetZoom` I think it will end up deleting the options for that scale. I think that we need to check if there are any new scales in `afterUpdate`. Once that's done you'll no longer need to store them in `afterInit` since they'll be getting stored in `afterUpdate`

jannisgt:
So I just tested it. It worked. Why do you think should resetZoom not cover this scenario?
The function `storeOriginalOptions` get called for every resetZoom/zoom/pan so it should work in each case.
But I found another case. When I delete a scale `y-axis-4` and add it again without reset/zoom/pan in between it does not check that it is a new one. But I think that is negligible

benmccann:
Oh, you're right. I missed that it was also called in `resetZoom`. I don't think we need the call in `afterInit`. It looks like it should be covered by the other cases. Can we remove that one?

Issue #61, opened 2017-01-30
Calling .resetZoom() on a chart destroys all metadata, losing more metadata than necessary in the process


AlbinoDrought:
When a user clicks a "legend square" to hide a line on the chart, ChartJS stores this in the metadata of the dataset.

When resetZoom() is called on a chart, it deletes all metadata. This causes, among other things, for the hidden metadata to be lost and the line to become visible again.

Example: http://codepen.io/albinodrought/pen/OWzqzY

### To Reproduce:
1. Hide a dataset by clicking on the legend "square"
2. Click the "Reset Zoom" link
3. Notice that the dataset hidden in step 1 is now visible again

AlbinoDrought:
I did some fiddling and noticed if I comment out the call that nulls metadata, this issue no longer occurs.

That call is the only part of the code I can see that directly accesses the metadata. I'm not sure of ChartJS internals, though, so there could be something going on in the background (possibly related to hiding/slicing of datapoints?).

mkoutli:
Hello, where did you comment out the call that nulls metadata? And have you faced problems since _meta was not cleared anymore?

AlbinoDrought:
@mkoutli https://github.com/chartjs/chartjs-plugin-zoom/blob/master/src/chart.zoom.js#L307

PR #204, opened 2019-02-23
Don't delete dataset._meta

benmccann:
Fixes https://github.com/chartjs/chartjs-plugin-zoom/issues/61

I tested this fix manually against `zoom-time.html` 

I'm guessing the reason it works now but the code was originally required is because the code no longer does `JSON.parse(JSON.stringify(scale.options))` and now does `helpers.clone` instead:

https://github.com/chartjs/chartjs-plugin-zoom/blob/f779867fe2dc2aa66bbdc0959df664dda9461b69/src/plugin.js#L329

### DIFF
--- a/samples/zoom-time.html
+++ b/samples/zoom-time.html
@@ -85,7 +85,7 @@
 					xAxes: [{
 						type: 'time',
 						time: {
-							format: timeFormat,
+							parser: timeFormat,
 							// round: 'day'
 							tooltipFormat: 'll HH:mm'
 						},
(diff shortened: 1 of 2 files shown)


jledentu:
@benmccann I tested but encountered an issue: in `zoom-time.html`, when I click on a legend item, all points are then displayed on the same X coordinate. I need to click on "Reset zoom" to display the chart correctly.

I can reproduce this issue on master, so I guess this PR is OK.

benmccann:
Ok. I think I figured out what's happening: https://github.com/chartjs/chartjs-plugin-zoom/issues/205

I agree it seems unrelated. If you're otherwise happy with this PR and are able to approve it then I'll go ahead and merge this

Issue #187, opened 2018-12-18
problem with  rangeMax and Min in zoom


DiruzCode:
Hi, 
When i try to use the rangeMax and Min in the property zoom, this not work, i don't know for what, but if use the same in the pan property it works

```
                 pan: { 
                   enabled: true,
                       mode: 'x',`
                       rangeMax: {
                      			x: new Date(),
                      			y: null
                       },
                       rangeMin: {
                      			x: new Date('2017-12-10'),
                      			y: null
                       }
                   },
                   zoom: {
                       enabled: true,
                       mode: 'x'
                       rangeMax: {
                      	 		x: new Date(),
                      			y: 40000
                        },
                        rangeMin: {
                       			x: new Date('2017-12-10'),
                       			y: 0
                        }
                   }
```

<img width="1186" alt="captura de pantalla 2018-12-18 a la s 17 59 20" src="https://user-images.githubusercontent.com/12041331/50182653-abd75a00-02ee-11e9-951d-284be1d2de19.png">

<img width="1157" alt="captura de pantalla 2018-12-18 a la s 18 02 28" src="https://user-images.githubusercontent.com/12041331/50182827-1f796700-02ef-11e9-91bb-1b4e2f09c6ad.png">

Any can help me? 

Regarts

benmccann:
Please format code blocks by placing three backticks before the whole block and three after

redbmk:
For me it seems like `rangeMax` works properly (doesn't allow me to zoom out past that date), but adding a `rangeMin` makes it so zoom doesn't work at all anymore. I've tried both `Date` objects as well as unix timestamps.

benmccann:
You need to pass unix timestamps. But even then it's broken. I sent a PR to fix here: https://github.com/chartjs/chartjs-plugin-zoom/pull/208

PR #208, opened 2019-03-01
Share zoom logic between time and numerical scales

benmccann:
Closes https://github.com/chartjs/chartjs-plugin-zoom/issues/187

It looks like the wrong variable was being passed to `rangeMinLimiter` and `rangeMaxLimiter` for the time scale. This fixes it by sharing the logic for the numerical scale

### DIFF
--- a/samples/zoom-time.html
+++ b/samples/zoom-time.html
@@ -24,6 +24,7 @@
 	</div>
 	<script>
 		var timeFormat = 'MM/DD/YYYY HH:mm';
+		var now = window.moment();
 
 		function randomScalingFactor() {
 			return Math.round(Math.random() * 100 * (Math.random() > 0.5 ? -1 : 1));
@@ -38,11 +39,11 @@
 		}
 
 		function newDate(days) {
-			return window.moment().add(days, 'd').toDate();
+			return now.clone().add(days, 'd').toDate();
 		}
 
 		function newDateString(days) {
-			return window.moment().add(days, 'd').format(timeFormat);
+			return now.clone().add(days, 'd').format(timeFormat);
 		}
 
 		var config = {
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -114,44 +114,12 @@ function zoomCategoryScale(scale, zoom, center, zoomOptions) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


anxorl:
rangeMax and rangeMin don't work for zoom after this change. Zoom goes out by several years.
Maybe, implementation analogous to panTimeScale is not the right choice.

benmccann:
I just tested and it worked for me. Perhaps there's something different about your code than mine

Techn1x:
This commit has broken my use of `onZoom`

Previously I was doing this..
```javascript
zoom: {
  ...
  mode: 'x',
  onZoom: function({chart}) { 
    // Pass the new required min/max time to the telemetry task
    let msStart = chart.scales['timeAxis'].min; // new min value eg 1556031945765
    let msEnd = chart.scales['timeAxis'].max; // new max value eg 1556067585765
    ...
    telemetryAction(msStart,msEnd);
  },
  rangeMin: {
    x: 60000 // 60 seconds
  },
  rangeMax: {
    x: 39600000 // 11 hours
  }
}
```

Now the values returned are different
```
let msStart = chart.scales['timeAxis'].min; // new min value eg 1556031969255
let msEnd = chart.scales['timeAxis'].max; // rangeMax value eg 39600000
```

I can no longer get the new min/max value after a zoom in. The scale `max` always gives the max range, and isn't updated. I can't figure out any way to get the updated zoom min/max information from the chart instance...

benmccann:
@Techn1x so the zoom behaves properly from a UI perspective, but you just can't retrieve the values during `onZoom`?

```
let msStart = chart.scales['timeAxis'].min; // new min value eg 1556031969255
let msEnd = chart.scales['timeAxis'].max; // rangeMax value eg 39600000
```
That seems surprising because it shouldn't be possible for `max` to be less than `min`: https://github.com/chartjs/Chart.js/blob/master/src/scales/scale.time.js#L581

>I can't figure out any way to get the updated zoom min/max information (the time diff) from the chart instance...

Did you also try `scale.options.time.min` / `scale.options.ticks.min`?

I don't have much time to debug this, but if you're able to debug it yourself I can help with getting a fix merged

PR #214, opened 2019-03-05
Version 0.7.0

benmccann:
Release notes here: https://github.com/chartjs/chartjs-plugin-zoom/releases/edit/untagged-cef6534833172b3552b3

@simonbrunel I think you will need to create the release branch for us

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,15 +3,21 @@
 A zoom and pan plugin for Chart.js. Currently requires Chart.js >= 2.6.0
 
 Panning can be done via the mouse or with a finger.
-Zooming is done via the mouse wheel or via a pinch gesture. [Hammer.js](http://hammerjs.github.io/) is used for gesture recognition.
+Zooming is done via the mouse wheel or via a pinch gesture. [Hammer.js](https://hammerjs.github.io/) is used for gesture recognition.
 
-[Live Codepen Demo](http://codepen.io/pen/PGabEK)
+[Live Codepen Demo](https://codepen.io/pen/PGabEK)
 
 ## Installation
 
 Run `npm install --save chartjs-plugin-zoom` to install with `npm`.
 
-If including via a `<script>` tag, make sure to include `Hammer.js` as well.
+If including via a `<script>` tag, make sure to include `Hammer.js` as well:
+
+```html
+<script src="https://cdn.jsdelivr.net/npm/chart.js@2.8.0"></script>
+<script src="https://cdn.jsdelivr.net/npm/hammerjs@2.0.8"></script>
+<script src="https://cdn.jsdelivr.net/npm/chartjs-plugin-zoom@0.7.0"></script>
+```
 
 ## Configuration
 
@@ -97,7 +103,7 @@ Prior to v0.4.0, this plugin was known as 'Chart.Zoom.js'. Old versions are stil
 
 ## Documentation
 
-You can find documentation for Chart.js at [www.chartjs.org/docs](http://www.chartjs.org/docs).
+You can find documentation for Chart.js at [www.chartjs.org/docs](https://www.chartjs.org/docs).
 
 Examples for this plugin are available in the [samples folder](samples).
 
@@ -107,4 +113,4 @@ Before submitting an issue or a pull request to the project, please take a momen
 
 ## License
 
-chartjs-plugin-zoom.js is available under the [MIT license](http://opensource.org/licenses/MIT).
+chartjs-plugin-zoom.js is available under the [MIT license](https://opensource.org/licenses/MIT).
--- a/README.md
+++ b/README.md
@@ -3,9 +3,15 @@
 A zoom and pan plugin for Chart.js. Currently requires Chart.js >= 2.6.0
 
 Panning can be done via the mouse or with a finger.
-Zooming is done via the mouse wheel or via a pinch gesture. [Hammer JS](http://hammerjs.github.io/) is used for gesture recognition.
+Zooming is done via the mouse wheel or via a pinch gesture. [Hammer.js](http://hammerjs.github.io/) is used for gesture recognition.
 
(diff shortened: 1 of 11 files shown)


simonbrunel:
Is it ready to be released?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "chartjs-plugin-zoom",
   "description": "Simple HTML5 charts using the canvas element.",
```
simonbrunel: Should we update the package description?

package.json
```diff
@@ -1,7 +1,7 @@
...
-  "version": "0.6.6",
+  "version": "0.7.0",
   "author": "Evert Timberg <evert.timberg@gmail.com>",
```
simonbrunel: I would remove `author` as we did in the core repo and a few plugins

simonbrunel:
I think `chart.js` should be a peer dependency as in the [datalabels plugin](https://github.com/chartjs/chartjs-plugin-datalabels/blob/master/package.json#L54) but it may be a breaking change if anyone installed this plugin only, but not `chart.js` explicitly (thus don't have `chart.js` as dependency of their projects `package.json`).

package.json
benmccann: done

package.json
benmccann: done

jledentu:
@benmccann Don't you think that the *Breaking Changes* section should be above other sections, in the release note? These breaking changes are really important.

benmccann:
I moved _Breaking Changes_ to the top

I also kept Hammer.js as an external dependency and added it to the README. The thing I hadn't realized earlier is that the `main` entry point in the `package.json` file was a bundled file that included Hammer.js. This means that if a user includes Hammer.js elsewhere in their application there's no opportunity for the bundler to dedup the dependency.

Let's wait a little bit on the release to get https://github.com/chartjs/chartjs-plugin-zoom/pull/215 merged first

README.md
```diff
@@ -3,10 +3,16 @@
...
+Zooming is done via the mouse wheel or via a pinch gesture. [Hammer.js](http://hammerjs.github.io/) is used for gesture recognition.
 
 [Live Codepen Demo](http://codepen.io/pen/PGabEK)
 
+## Installation
+
+Run `npm install --save chartjs-plugin-zoom` to install with `npm`.
+
+If including via a `<script>` tag, make sure to include `Hammer.js` as well.
+
```
simonbrunel: I would add an example:

```
If including via a `<script>` tag, make sure to include `Hammer.js` as well:

```html
<script src="https://cdn.jsdelivr.net/npm/chart.js@2.8.0"></script>
<script src="https://cdn.jsdelivr.net/npm/hammerjs@2.0.8"></script>
<script src="https://cdn.jsdelivr.net/npm/chartjs-plugin-zoom@0.7.0"></script>
```

Hammer doesn't need the extra path to the min file (`/hammer.min.js`). Until Chart.js 2.8, the path to the dist file was required (`chart.js@2.7.3/dist/Chart.min.js`) but it's not anymore the case since we changed [entry points](https://github.com/chartjs/Chart.js/blob/master/package.json#L7).

We may want the same for this plugin:

```javascript
"jsdelivr": "dist/chartjs-plugin-zoom.min.js",
"unpkg": "dist/chartjs-plugin-zoom.min.js",
"main": "dist/chartjs-plugin-zoom.js",
```

README.md
```diff
@@ -3,10 +3,16 @@
 A zoom and pan plugin for Chart.js. Currently requires Chart.js >= 2.6.0
 
 Panning can be done via the mouse or with a finger.
-Zooming is done via the mouse wheel or via a pinch gesture. [Hammer JS](http://hammerjs.github.io/) is used for gesture recognition.
+Zooming is done via the mouse wheel or via a pinch gesture. [Hammer.js](http://hammerjs.github.io/) is used for gesture recognition.
```
simonbrunel: `https://hammerjs....`

samples/pan-bar.html
```diff
@@ -3,9 +3,9 @@
 
 <head>
 	<title>Bar Chart Pan</title>
-	<script src="../node_modules/chart.js/dist/Chart.bundle.js"></script>
-	<script src="../node_modules/hammerjs/hammer.min.js"></script>
-	<script src="../dist/chartjs-plugin-zoom.js"></script>
+	<script src="https://cdn.jsdelivr.net/npm/chart.js@2.7.3/dist/Chart.min.js"></script>
+	<script src="https://cdn.jsdelivr.net/npm/hammerjs@2.0.8/hammer.min.js"></script>
```
simonbrunel: That are the links I sent you on Slack but actually we don't need the extra file path for Hammer. It will be the same with Chart.js starting v2.8, so this would become:

```html
<script src="https://cdn.jsdelivr.net/npm/chart.js@2.8.0"></script>
<script src="https://cdn.jsdelivr.net/npm/hammerjs@2.0.8"></script>
<script src="../dist/chartjs-plugin-zoom.min.js"></script>
```

package.json
```diff
@@ -29,8 +28,10 @@
     "rollup-plugin-terser": "^4.0.2",
     "semver": "^5.6.0"
   },
+  "peerDependencies": {
+    "chart.js": "^2.6.0"
+  },
   "dependencies": {
-    "chart.js": "^2.6.0",
     "hammerjs": "^2.0.8"
```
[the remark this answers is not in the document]
benmccann: The goal was to have it specified as a dependency so that npm would pull it in. We had to make it external so that it wouldn't be bundled and pulled in via npm. I think npm should manage the dependency so that if another dependency includes hammer it can be deduped. I don't think the user should have to specify hammer directly in their project though. So I do think `dependencies` is probably the right place to specify it

README.md
benmccann: done. updated everywhere else as well

PR #215, opened 2019-03-05
Add speed option

### LINKED ISSUE #50: Zoom speed?
Hi, I have tried varying Zoom sensitivity value from 10 (very slow) to 0.0001 (pretty fast) but it seems that the dataset is Zoomed in 1 data point at a time.

Is there any way to control the 'step' of zoom function?

Thanks for any help.

jledentu:
This PR adds a `speed` option for `zoom`. This option allows to configure the percentage of zoom applied on a wheel event.

Closes #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ To configure the zoom and pan plugin, you can simply add new config options to y
 
 		// Speed of zoom via mouse wheel
 		// (percentage of zoom on a wheel event)
-		speed: 10,
+		speed: 0.1,
 
 		// Function called once zooming is completed
 		// Useful for dynamic data loading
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -64,6 +65,11 @@ To configure the zoom and pan plugin, you can simply add new config options to y
 			x: null,
 			y: null
 		},
+
+		// Speed of zoom via mouse wheel
+		// (percentage of zoom on a wheel event)
+		speed: 10,
```
benmccann: I probably would have gone with `0.1` since it's a percentage

jledentu: Yeah I wasn't sure about what was the most intuitive... I changed that.

Issue #195, opened 2019-01-25
Check config in a correct place too (options.plugins)


PaulAnnekov:
Currently this plugin waits for config in `options.zoom` and `options.pan` while the correct place for them is `options.plugins.zoom` and `options.plugins.pan` respectively (https://www.chartjs.org/docs/latest/developers/plugins.html#plugin-options). This problem also makes a pain to use the plugin with TypeScript, because it doesn't expect `options` to contain `zoom` and `pan` properties and emits an error.

jledentu:
You're right. I suggest to handle this in the future 1.0 version as it would break the API. @benmccann We may create a 1.0 milestone to reference issues with breaking changes?

benmccann:
We could read from both locations now and remove the old one at 1.0. We could add a `TODO(1.0)` in the code to track in lieu of an issue

simonbrunel:
> @PaulAnnekov This problem also makes a pain to use the plugin with TypeScript

FYI, DefinitelyTyped/DefinitelyTyped#31440 now allows to extend the TypeScript definition for the plugin options. That means plugins can *independently* provide definition files for their public API. For example, the datalabels plugin now exposes these types in the [`/types` folder](/chartjs/chartjs-plugin-datalabels/tree/master/types) ([`9e0026597`](/chartjs/chartjs-plugin-datalabels/commit/9e0026597a618f7ab1b07f04c75bc6395f96bbf6)).

benmccann:
@PaulAnnekov a PR would be welcome for this

benmccann:
This has been fixed

PR #217, opened 2019-03-06
Pan and zoom can now be configured from root or plugins

ezpzlmnsqz1337:
The plugin can now accept configuration options directly from options, or from a options.plugins.zoom.

Related to this issue #195 

I also copied 3 of the samples, to show how to configure it using the plugins object.


benmccann:
Thanks for this @ezpzlmnsqz1337. I don't think we need new samples. I would suggest just editing the existing samples. It's probably best to update all the examples so that we demonstrate the correct configuration

jledentu:
@ezpzlmnsqz1337 Great! I think you should update the `README` to document this `options.plugins.zoom` property (as replacement for the legacy options structure).

ezpzlmnsqz1337:
Ok, I will change the existing examples and README

ezpzlmnsqz1337:
So I updated the samples and readme, but I see that in zoom.html, after recreating the chart, an issue occurs.

```
Chart.bundle.js:11925 Uncaught TypeError: Cannot set property 'font' of null
    at ChartElement.calculateTickRotation (Chart.bundle.js:11925)
    at ChartElement.update (Chart.bundle.js:11833)
    at getMinimumBoxSize (Chart.bundle.js:10974)
    at Object.each (Chart.bundle.js:14458)
    at Object.update (Chart.bundle.js:10985)
    at Chart.updateLayout (Chart.bundle.js:8785)
    at Chart.update (Chart.bundle.js:8740)
    at doZoom (chartjs-plugin-zoom.js:2853)
    at HTMLCanvasElement.chartInstance.zoom._wheelHandler (chartjs-plugin-zoom.js:3120)
```

I'll try to fix it, but currently I have no clue what is causing it. Any idea is welcome.

ezpzlmnsqz1337:
Ok guys, so both issues addressed above are fixed. It should be all working fine now.

src/plugin.js
```diff
@@ -160,13 +160,14 @@ function doZoom(chart, percentZoomX, percentZoomY, focalPoint, whichAxes) {
 		};
 	}
 
-	var zoomOptions = chart.options.zoom;
+	var zoomOptions = chart.options.zoom || chart.options.plugins.zoom.zoom;
```
simonbrunel: Check https://github.com/chartjs/chartjs-plugin-zoom/issues/195#issuecomment-458997325: plugin options shouldn't be read directly from the chart options but retrieved from the [hooks last argument](https://github.com/chartjs/Chart.js/blob/master/src/core/core.plugins.js#L182) ([example](/chartjs/chartjs-plugin-datalabels/blob/9e0026597a618f7ab1b07f04c75bc6395f96bbf6/src/plugin.js#L120)). That allows to handle some extra cases, for example if the user set `options.plugins.zoom: true`, you will receive the default options.

README.md
```diff
@@ -9,7 +9,71 @@ Zooming is done via the mouse wheel or via a pinch gesture. [Hammer JS](http://h
...
+			},
+			// Function called once zooming is completed
+			// Useful for dynamic data loading
+			onZoom: function({chart}) { console.log(`I was zoomed!!!`); }
+		}
+	}
+}
+```
+
+For compatibility reasons it is possible to set these options directly on the `chart.options` object, but putting it into `chart.options.plugins.zoom` object is the right way.
```
benmccann: Let's not include all the options again after this line. Demonstrating them above is enough

Also, minor wording suggestion. How about:

To maintain compatibility with previous version of the plugin, it is possible to set these options directly on the `chart.options` object. However, `chart.options.plugins.zoom` is the preferred place to specify them.

src/plugin.js
benmccann: Another thing I'm wondering. We might want to include a warning to help people migrate like::

```
var options = chartInstance.options.zoom;
if (options) {
    console.warn('Please specify chartjs-plugin-zoom options via options.plugin.zoom');
} else {
    options = pluginOptions;
}
```

README.md
ezpzlmnsqz1337: Removed second settings, replaced with your wording.

src/plugin.js
ezpzlmnsqz1337: Ok so I changed that in the hooks, and it works. Only problem is, that in beforeInit you are registering the events, and I can't use the pluginOptions from the hook, because if user changes the options, the event is still using the old options that it was created with. So there I left it using the options.plugins.zoom. If you know how to get around this, tell me.

ezpzlmnsqz1337: Also I added the warning in beforeInit hook.

samples/zoom-time.html
```diff
@@ -127,7 +131,7 @@
 		};
 
 		window.toggleDragMode = function() {
-			var zoomOptions = window.myLine.options.zoom;
+			var zoomOptions = window.myLine.options.zoom || window.myLine.options.plugins.zoom.zoom;
```
benmccann: This can just be:

    var zoomOptions = window.myLine.options.plugins.zoom.zoom

src/plugin.js
benmccann: Hmmm. Did you test this? It doesn't look to me like `beforeInit` passes the options. I'm also worried there are other places where we're going to be using un-updated options

```
Chart.js $ grep -r beforeInit src/
src/core/core.plugins.js: * @method IPlugin#beforeInit
src/core/core.controller.js:		plugins.notify(me, 'beforeInit');
src/plugins/plugin.legend.js:	beforeInit: function(chart) {
src/plugins/plugin.title.js:	beforeInit: function(chart) {
```

You might have to revert that change. Thoughts @simonbrunel ?

simonbrunel: Your `grep` isn't pertinent: the `title` and `legend` plugins don't use "plugins" options (for historical reasons). Also, options are [always passed as the last argument](/chartjs/Chart.js/blob/master/src/core/core.plugins.js#L108), so code calling `plugins.notify()` doesn't have to care about it. Here is an [example where `options` is used in the `beforeInit` hook](/chartjs/chartjs-plugin-deferred/blob/master/src/plugin.js#L131).

benmccann: Ah, I missed where options is passed as the last argument. In that case it means it's also passed to `afterUpdate`, so we should add a hook for `afterUpdate` as well to store the updated options. Right now there are cases in this PR where we're only using the initial options

ezpzlmnsqz1337: I tried the examples and all seem to be working. The only issue was, that in the beforeInit hook, are event registrations like so:
```
chartInstance.zoom._mouseMoveHandler = function(event) {
	var opts = chartInstance.options;
	var zoomOpts = opts.zoom || opts.plugins.zoom.zoom;
	if (zoomOpts && zoomOpts.drag && chartInstance.zoom._dragZoomStart) {
		chartInstance.zoom._dragZoomEnd = event;
		chartInstance.update(0);
	}
};
```
If in this event registration I pass the options from the beforeInit hook, and later in the web page I change the value of "drag" (this is in the `zoom-time.html` sample) the event will not get the updated options.

However passing the options from the beforeInit hook to the method "doZoom" works correctly and reacts to changes. I am not really sure now why it works with a method call and not within event registration.

samples/zoom-time.html
ezpzlmnsqz1337: ok, changed.

src/plugin.js
benmccann: I cannot figure out what's happening here. If I toggle the drag setting in `samples/zoom-time.html` and set a breakpoint in an event handler in `beforeInit` then I see `chartInstance.$plugins.descriptors[3].options` is updated, but `pluginOptions` is not. However, it looks to me like they should be the same object, so I'm not sure why their values would be different. Any ideas @simonbrunel ?

benmccann:
Closing this in favor of https://github.com/chartjs/chartjs-plugin-zoom/pull/221

PR #221, opened 2019-03-08
Cleanup options handling

benmccann:
Closes https://github.com/chartjs/chartjs-plugin-zoom/pull/217 & https://github.com/chartjs/chartjs-plugin-zoom/issues/195

FYI @ezpzlmnsqz1337

@simonbrunel pointed out several issues with the plugin's current handling of options in the #dev channel on Slack. This attempts to address those issues in a more holistic manner

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -12,7 +12,7 @@ var zoomNS = Chart.Zoom = Chart.Zoom || {};
 var zoomFunctions = zoomNS.zoomFunctions = zoomNS.zoomFunctions || {};
 var panFunctions = zoomNS.panFunctions = zoomNS.panFunctions || {};
 
-Chart.defaults.global.plugins.zoom = {
+Chart.Zoom.defaults = Chart.defaults.global.plugins.zoom = {
 	pan: {
 		enabled: true,
 		mode: 'xy',
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -12,8 +12,7 @@ var zoomNS = Chart.Zoom = Chart.Zoom || {};
 var zoomFunctions = zoomNS.zoomFunctions = zoomNS.zoomFunctions || {};
 var panFunctions = zoomNS.panFunctions = zoomNS.panFunctions || {};
 
-// Default options if none are provided
-var defaultOptions = zoomNS.defaults = {
+Chart.defaults.global.plugins.zoom = {
 	pan: {
 		enabled: true,
 		mode: 'xy',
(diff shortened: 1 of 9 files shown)


benmccann:
@jledentu just a friendly reminder to review this when you get a chance. We can release 0.7.0 afterwards

src/plugin.js
```diff
@@ -12,8 +12,7 @@ var zoomNS = Chart.Zoom = Chart.Zoom || {};
 var zoomFunctions = zoomNS.zoomFunctions = zoomNS.zoomFunctions || {};
 var panFunctions = zoomNS.panFunctions = zoomNS.panFunctions || {};
 
-// Default options if none are provided
-var defaultOptions = zoomNS.defaults = {
+Chart.defaults.global.plugins.zoom = {
```
simonbrunel: I think users used to configure defaults via `Chart.Zoom.defaults`, so I think you need to keep it working until v1.

```javascript
Chart.Zoom.defaults = Chart.defaults.global.plugins.zoom = {
  // ...
}
```

benmccann: done

jledentu:
@benmccann Just tested. It works with the current structure of options. But I don't understand why this PR replaces #217 which addresses another issue (handling `chart.options.plugins.zoom`).

I tested a sample with: 

```js
plugins: {
	zoom: {
		pan: {
			enabled: true,
			mode: 'xy' // is panning about the y axis neccessary for bar charts?
		},
		zoom: {
			enabled: true,
			mode: 'x',
			sensitivity: 3
		},
	}
},
```

and it doesn't work?

benmccann:
Thanks for testing @jledentu. It's meant to handle both the old and new cases. I fixed the issue, so hopefully it should work for you now. Thanks for catching that!

jledentu:
The `Disable drag mode` doesn't work anymore in `zoom-time.html`, you have to replace:

```js
var zoomOptions = window.myLine.options.zoom;
```

by:

```js
var zoomOptions = window.myLine.options.plugins.zoom.zoom;
```

I tried to do that but the change in options doesn't seem to be taken into account immediately? :thinking:

benmccann:
Good catch. Yeah, Chart.js usually makes you call `update()` after updating the options. This plugin was an exception. Now that we're using more of the built-in options functionality you'll need to call `update()` after updating the options. That aligns more with the main Chart.js repo, but we'll have to be sure to mention it in the release notes

benmccann:
Thanks for the great testing on this one @jledentu !

PR #231, opened 2019-04-16
Fix drag zoom if either x or y directions are disabled

### LINKED ISSUE #205: Single click on legend is interpreted as drag
Clicking on the legend activates the drag feature. If possible, only clicking within the chart area should activate this feature

Additionally, a single click anywhere on the chart is interpreted as a drag. It perhaps should not be unless the mouse moves.

See this in the zoom-time.html sample

troygibb:
This PR fixes #205. Pretty straightforward bug, when you only have `x` or `y` directions enabled, `doZoom` is still called on single clicks for `drag` configs since it interprets `dragDistanceY` either `dragDistanceX` as truthy.

I tested all iterations on the `/samples/zoom-time.html` example, and can confirm that the current version has the bug.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -449,8 +449,9 @@ var zoomPlugin = {
 			// Remove drag start and end before chart update to stop drawing selected area
 			chartInstance.$zoom._dragZoomStart = null;
 			chartInstance.$zoom._dragZoomEnd = null;
-
-			if (dragDistanceX > 0 || dragDistanceY > 0) {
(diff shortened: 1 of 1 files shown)


benmccann:
Thanks for looking at this Troy and tracking down the issue

It seems to me like `dragDistanceX` and `dragDistanceY` should both be 0 and this change is sort of working around the fact that they're not. What do you think about https://github.com/chartjs/chartjs-plugin-zoom/pull/232 as another solution?

troygibb:
@benmccann ah yes I was thinking about that kind of solution too--I'll close in favor. (Verified your change locally as well).

PR #232, opened 2019-04-17
Fix calculation of dragDistance for click

benmccann:
Closes https://github.com/chartjs/chartjs-plugin-zoom/issues/205

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -418,33 +418,25 @@ var zoomPlugin = {
 			node.removeEventListener('mousemove', chartInstance.$zoom._mouseMoveHandler);
 
 			var chartArea = chartInstance.chartArea;
-			var xAxis = getXAxis(chartInstance);
-			var yAxis = getYAxis(chartInstance);
 			var beginPoint = chartInstance.$zoom._dragZoomStart;
-			var startX = xAxis.left;
-			var endX = xAxis.right;
-			var startY = yAxis.top;
-			var endY = yAxis.bottom;
 
-			if (directionEnabled(chartInstance.$zoom._options.zoom.mode, 'x')) {
-				var offsetX = beginPoint.target.getBoundingClientRect().left;
-				startX = Math.min(beginPoint.clientX, event.clientX) - offsetX;
-				endX = Math.max(beginPoint.clientX, event.clientX) - offsetX;
-			}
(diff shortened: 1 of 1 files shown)


troygibb:
@benmccann any chance this will be merged any time soon?

benmccann:
I don't have the ability to merge it until someone reviews it

src/plugin.js
```diff
@@ -418,33 +418,25 @@ var zoomPlugin = {
...
 			var chartArea = chartInstance.chartArea;
-			var xAxis = getXAxis(chartInstance);
-			var yAxis = getYAxis(chartInstance);
 			var beginPoint = chartInstance.$zoom._dragZoomStart;
-			var startX = xAxis.left;
-			var endX = xAxis.right;
-			var startY = yAxis.top;
-			var endY = yAxis.bottom;
 
-			if (directionEnabled(chartInstance.$zoom._options.zoom.mode, 'x')) {
```
simonbrunel: The previous logic seems better since we avoid calculating extra values in case `directionEnabled` was `false`

benmccann: The previous logic was buggy, so hardly better :-)

I sent a follow up that's optimized better than the previous version even: https://github.com/chartjs/chartjs-plugin-zoom/pull/236

I don't think the change will be meaningfully different, but it doesn't hurt either, so I'm not opposed to it

simonbrunel: Obviously, I didn't mean the whole previous logic was better :)

`offsetX`, `startX`, `endX` and `dragDistanceX` (same for the `Y` variants) don't need to be calculated if `xEnabled` (respectively `yEnabled`) are `false` (still the case in the new code).

benmccann: Hmm. `dragDistanceX` is used here: https://github.com/chartjs/chartjs-plugin-zoom/blob/master/src/plugin.js#L447. That parameter ends up being ignored in the case that `xEnabled=false`, but I'm not sure I like the idea of passing in a junk value knowing what that method will do internally. And I'd rather not spend developer time worrying about saving a couple subtraction operations which have no real cost

Attrash-Islam:
Is there a chance to publish a new version on npm that includes this fix?

troygibb:
Thanks again @benmccann for getting this in!!!

PR #237, opened 2019-05-11
Fix #233: onPan function is called multiple times (and not as stated …

### LINKED ISSUE #233: onPan method calls 6-10 times on ajax request when dragging
```
plugins: {
    zoom: {
        pan: {
            enabled: true,
            mode: 'x',
            onPan: function(event) {
                
                self.refreshChartLine(function (xAxes, yAxes, color) {
                    self.plot([{
                        x: xAxes,
                        y: yAxes,
                        color: color
                    }]);
                });
            }
        },
        zoom: {
            enabled: true,
            drag: false,
            mode: 'x'
        }
    }
}
```

### Is there way how to do it correctly? Example: one dragging = one request

maxstrauch:
See also Issue #233 (https://github.com/chartjs/chartjs-plugin-zoom/issues/233).

Currently the documentation ([README.md](https://github.com/chartjs/chartjs-plugin-zoom/blob/master/README.md)) says:

    // Function called once panning is completed
    // Useful for dynamic data loading
    onPan: function({chart}) { console.log(`I was panned!!!`); }

which is not correct. In `src/plugin.js` it can clearly be seen:

    var handlePan = function(e) {
        ...
            doPan(chartInstance, deltaX, deltaY);
        ...
    };

    mc.on('panstart', function(e) {
        ...
        handlePan(e);
    });
    mc.on('panmove', handlePan);
    mc.on('panend', function() {
        ...
    });

    // ...

    function doPan(chartInstance, deltaX, deltaY) {
        ...
    		if (typeof panOptions.onPan === 'function') {
    			panOptions.onPan({chart: chartInstance});
    		}
    	...
    }

That the function is called on `panstart` and `panmove`. For the intended use-case (e.g. loading dynamically data) this is not helpful at all.

Therefore this merge request fixes the issue and adds besides the `onPan` callback also a `onPanChange` callback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,12 @@ plugins: {
 				x: null,
 				y: null
 			},
-			// Function called once panning is completed
+            // Function called while the user is panning
+            // Useful for internal model state update
+            onPan: function({chart}) { console.log(`I'm panning!!!`); }
+            // Function called once panning is completed
 			// Useful for dynamic data loading
-			onPan: function({chart}) { console.log(`I was panned!!!`); }
+            onPanComplete: function({chart}) { console.log(`I was panned!!!`); }
 		},
 
 		// Container for zoom options
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -291,8 +291,8 @@ function doPan(chartInstance, deltaX, deltaY) {
 
 		chartInstance.update(0);
 
-		if (typeof panOptions.onPan === 'function') {
-			panOptions.onPan({chart: chartInstance});
+		if (typeof panOptions.onPanChange === 'function') {
+			panOptions.onPanChange({chart: chartInstance});
 		}
 	}
 }
@@ -558,7 +558,12 @@ var zoomPlugin = {
 				zoomNS.panCumulativeDelta = 0;
 				setTimeout(function() {
 					panning = false;
-				}, 500);
+                }, 500);
+                
+                var panOptions = chartInstance.$zoom._options.pan;
+                if (typeof panOptions.onPan === 'function') {
+                    panOptions.onPan({chart: chartInstance});
+                }
 			});
 
 			chartInstance.$zoom._ghostClickHandler = function(e) {
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -291,8 +291,8 @@ function doPan(chartInstance, deltaX, deltaY) {
 
 		chartInstance.update(0);
 
-		if (typeof panOptions.onPanChange === 'function') {
-			panOptions.onPanChange({chart: chartInstance});
+		if (typeof panOptions.onPan === 'function') {
+			panOptions.onPan({chart: chartInstance});
 		}
 	}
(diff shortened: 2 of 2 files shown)


benmccann:
Thanks. What do you think about `onPan` and `onPanComplete` as names?

benmccann:
It might be good to add the same functionality for zoom at the same time for consistency

maxstrauch:
@benmccann I also thought afterwards on `onPan` and `onPanComplete`, because it does not break current apps, relying on the behaviour of `onPan`

arslanhashmi:
@benmccann can you please merge this? I also need these callbacks to work well in my project.

maxstrauch:
@benmccann @arslanhashmi I finally got around it and updated my fix so that `onPan` remains how it is now (called multiple times) and does not break existing applications which rely on this. For the final callback I named it `onPanComplete`. This should work fine.
I also adapted the `README.md` with the newest behaviour.

I think we can merge this now.

arslanhashmi:
@maxstrauch great is it same for zoom?

maxstrauch:
Not yet. I'm just looking into it right now to see if I can integrate it very easily.

arslanhashmi:
I hope this will solve few issues, as currently there is no context is passed with the callbacks: onZoom/onPan

maxstrauch:
@benmccann @arslanhashmi I think this fix cannot be easily implemented for `onZoom` since there is only an DOM event for `onwheel` but no event like e.g. `onwheelstop` (see https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onwheel).

However one could implement a workaround using `setTimeout` and assume that after e.g. 50ms of no action one distinct zoom cycle is finished. But I think this not a very nice solution since it is a workaround. Maybe it is better to use something like `rxjs` on the application side (which uses chartjs) and use `debounce` to process zoom events.

README.md
```diff
@@ -44,9 +44,12 @@ plugins: {
 				x: null,
 				y: null
 			},
-			// Function called once panning is completed
+            // Function called while the user is panning
```
benmccann: There's an extra space on most of these lines

maxstrauch:
@benmccann I think it would not hurt to just have both. Maybe I want to do some pre-loading while panning and commit it afterwards. I think it is better to have it in then to remove it. I don't think that this is a bloat-feature. **But you decide!**

For the other point: okay.

But one wish from me: could we please, please, please merge this? I really need it for a project.

benmccann:
I suppose I'd be fine to merge with the following two changes:
* If we're going in the direction of having `onPan` and `onPanComplete` then we should update `onZoom` documentation in the README to match its current functioning. 
* Also, see my comments about the spacing as that will need to be fixed before merging.

@jledentu do you have any thoughts about this PR?

benmccann:
Merged in https://github.com/chartjs/chartjs-plugin-zoom/pull/250

Issue #77, opened 2017-05-20
do not intercept scroll gestures when both pan and zoom are disabled


mingyuan-xia:
When both pan and zoom are disabled, HammerJS would still intercept gestures on the chart. Suppose a user is scrolling down the page with some charts. When the mouse encounters any chart, the page becomes non-scrollable.

BabOuDev:
+1, very annoying... will have a look on how to solve this!

BabOuDev:
solved for you, waiting for merge.  =)
https://github.com/chartjs/chartjs-plugin-zoom/pull/80

mingyuan-xia:
short but looks great. Thanks a lot. Waiting for the owner.

mingyuan-xia:
@compwright please release the patch to npmjs, thanks.

qxzcode:
This bug appears to be back in the current version on NPM (0.7.0). I'm having the same problem @mingyuan-xia was having: scrolling doesn't work on charts, even when both `pan` and `zoom` are disabled.

benmccann:
@jledentu I'm guessing this might be related to https://github.com/chartjs/chartjs-plugin-zoom/pull/196. Do you think you'd be able to take a look?

PR #244, opened 2019-06-21
Don't listen to 'wheel' event if zoom is not enabled

cooljeffro:
Fixes https://github.com/chartjs/chartjs-plugin-zoom/issues/77


benmccann:
I don't think this solution will work because it's only checking at chart init time. With this PR, you can no longer change the options to enable zoom after the chart is initially rendered.

We should come up with a solution that makes both #77 and #196 work.

PR #246, opened 2019-07-01
Add isModeEnabledFn to zoom options

davidswinegar:
## Summary
Add an optional function property, isModeEnabledFn to the zoom
options. This allows disabling modes based on other keyboard
properties.
Ex. disabling the 'y' direction when a particular key is pressed

## Explanation
👋 Sorry I did not add an issue ahead of time for this. We've been using this plugin in our project (it's great!) but there's one feature we needed that wasn't in this library. Please let me know what I should do to get this in the plugin - I'm open to suggestions on everything. The main use-case that we want to preserve is the ability to call a function that is passed in to decide whether to enable or disabling zooming in a particular direction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,13 @@ plugins: {
 			// Eg. 'y' would only allow zooming in the y direction
 			mode: 'xy',
 
+			// Optional function for disabling the enabled mode, given an event.
+			// Allows disabling modes based on other keyboard properties.
+			// Ex. disabling the 'y' direction when a particular key is pressed.
+			// event: nullable DOM event
+			// direction: 'x' or 'y'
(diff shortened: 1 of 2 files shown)


benmccann:
I'm a little confused because you talk about disabling the `y` direction when a particular key is pressed, but then the events passed to the callback are the pinch event or mouse up event.

In terms of API, I'd prefer to make mode take `string|function` since I think that's most comparable to how we handle other options like this in the main Chart.js library. And also probably to make the function take a single parameter that's an object that provides the context. It'd probably be useful to include the chart in that object as well

davidswinegar:
Closed in favor of #258

PR #258, opened 2019-07-15
Allow using a function for `mode`

### LINKED ISSUE #257: Allow disabling zoom/pan in a particular direction while panning or zooming
We've been using this plugin in our project (it's great!) but there's one feature we needed that wasn't in this library. We want the ability to call a function that is passed into the options to decide whether to enable or disabling zooming in a particular direction. PR to come.

davidswinegar:
Allow using a function for `mode` that takes the chart and returns the directions available to scroll. This allows disabling a scrollable x or y direction based on application state instead of a static option.

Fixes #257. Feedback taken from #246

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,10 @@ plugins: {
 
 			// Panning directions. Remove the appropriate direction to disable
 			// Eg. 'y' would only allow panning in the y direction
+			// A function that is called as the user is panning can also be used:
+			//   mode: function({chart, direction}: {chart: ChartJS.Chart, direction: 'x' | 'y'}) {
+			//     return direction === 'y'
+		    //   },
 			mode: 'xy',
 
 			rangeMin: {
@@ -69,6 +73,10 @@ plugins: {
 
 			// Zooming directions. Remove the appropriate direction to disable
 			// Eg. 'y' would only allow zooming in the y direction
+			// A function that is called as the user is panning can also be used:
+			//   mode: function({ chart, direction }: { chart: ChartJS.Chart, direction: 'x' | 'y' }) {
+			//     return direction === 'y'
+		    //   },
 			mode: 'xy',
 
 			rangeMin: {
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ plugins: {
 			// A function that is called as the user is panning and returns the
 			// available directions can also be used:
 			//   mode: function({ chart }) {
-			//     return 'xy'
+			//     return 'xy';
 			//   },
 			mode: 'xy',
 
@@ -77,7 +77,7 @@ plugins: {
 			// A function that is called as the user is zooming and returns the
 			// available directions can also be used:
 			//   mode: function({ chart }) {
-			//     return 'xy'
+			//     return 'xy';
 			//   },
 			mode: 'xy',
 
--- a/README.md
+++ b/README.md
@@ -33,9 +33,10 @@ plugins: {
 
 			// Panning directions. Remove the appropriate direction to disable
 			// Eg. 'y' would only allow panning in the y direction
-			// A function that is called as the user is panning can also be used:
-			//   mode: function({chart, direction}: {chart: ChartJS.Chart, direction: 'x' | 'y'}) {
-			//     return direction === 'y'
+			// A function that is called as the user is panning and returns the
+			// available directions can also be used:
+			//   mode: function({ chart }: { chart: ChartJS.Chart }) {
+			//     return 'xy'
 		    //   },
 			mode: 'xy',
 
@@ -73,9 +74,10 @@ plugins: {
 
 			// Zooming directions. Remove the appropriate direction to disable
 			// Eg. 'y' would only allow zooming in the y direction
-			// A function that is called as the user is panning can also be used:
-			//   mode: function({ chart, direction }: { chart: ChartJS.Chart, direction: 'x' | 'y' }) {
-			//     return direction === 'y'
+			// A function that is called as the user is zooming and returns the
+			// available directions can also be used:
+			//   mode: function({ chart }: { chart: ChartJS.Chart }) {
+			//     return 'xy'
 		    //   },
 			mode: 'xy',
 
--- a/README.md
+++ b/README.md
@@ -35,9 +35,9 @@ plugins: {
 			// Eg. 'y' would only allow panning in the y direction
 			// A function that is called as the user is panning and returns the
 			// available directions can also be used:
-			//   mode: function({ chart }: { chart: ChartJS.Chart }) {
+			//   mode: function({ chart }) {
 			//     return 'xy'
-		    //   },
+			//   },
 			mode: 'xy',
 
 			rangeMin: {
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -69,6 +73,10 @@ plugins: {
 
 			// Zooming directions. Remove the appropriate direction to disable
 			// Eg. 'y' would only allow zooming in the y direction
+			// A function that is called as the user is panning can also be used:
```
benmccann: This should say "zooming" instead of "panning"

ddeya:
Looking forward to this PM. I need this feature.

davidswinegar:
Yeah, I can make those changes.

davidswinegar:
@benmccann I updated this with those changes

README.md
```diff
@@ -33,9 +33,10 @@ plugins: {
 
 			// Panning directions. Remove the appropriate direction to disable
 			// Eg. 'y' would only allow panning in the y direction
-			// A function that is called as the user is panning can also be used:
-			//   mode: function({chart, direction}: {chart: ChartJS.Chart, direction: 'x' | 'y'}) {
-			//     return direction === 'y'
+			// A function that is called as the user is panning and returns the
+			// available directions can also be used:
+			//   mode: function({ chart }: { chart: ChartJS.Chart }) {
```
benmccann: The `{ chart }: { chart: ChartJS.Chart }` syntax is confusing to me and not consistent with the rest of the README. Can we make it just `{ chart }` instead? I think people will know it's a chart.js chart and `ChartJS` isn't really a defined namespace anyway

src/plugin.js
```diff
@@ -199,7 +204,7 @@ function doZoom(chart, percentZoomX, percentZoomY, focalPoint, whichAxes) {
 
 		// Which axe should be modified when figers were used.
 		var _whichAxes;
-		if (zoomMode === 'xy' && whichAxes !== undefined) {
+		if (directionEnabled(chart, zoomMode, 'x') && directionEnabled(chart, zoomMode, 'y') && whichAxes !== undefined) {
```
benmccann: Rather than pass `chart` into `directionEnabled`, I would resolve the option in `resolveOptions`. The reason I suggest that is that it'd be a bit more consistent with how we handle options in the main Chart.js library. You could perhaps do something like:

```
if (typeof options.zoom.mode === 'function') {
  options.zoom.mode = options.zoom.mode({chart: chart});
}
props._options = options;
```

src/plugin.js
```diff
@@ -37,6 +37,12 @@ function resolveOptions(chart, options) {
 	}
 	var props = chart.$zoom;
 	options = props._options = helpers.merge({}, [options, deprecatedOptions]);
+	if (typeof options.pan.mode === 'function') {
```
benmccann: Ah, this is a bit different than what I was suggesting. I see that you're doing it because you want to call the function as the user zooms or pans and not at `update` time. I missed that as a requirement you had originally. I'm not sure it makes sense to do in `resolveOptions` given that. Maybe a better place to do it would be when setting zoomMode:

https://github.com/chartjs/chartjs-plugin-zoom/blob/b5666c46e109119d1875fe3fc59f3b91bf6bf6e3/src/plugin.js#L209

    var zoomMode = zoomOptions.mode === 'function' ? zoomOptions.mode({chart: chartInstance}) : zoomOptions.mode;

src/plugin.js
```diff
@@ -199,7 +210,7 @@ function doZoom(chart, percentZoomX, percentZoomY, focalPoint, whichAxes) {
 
 		// Which axe should be modified when figers were used.
 		var _whichAxes;
-		if (zoomMode === 'xy' && whichAxes !== undefined) {
+		if (directionEnabled(zoomMode, 'x') && directionEnabled(zoomMode, 'y') && whichAxes !== undefined) {
```
benmccann: We need to support `panMode` on line 315 as well, right? 

https://github.com/chartjs/chartjs-plugin-zoom/blob/b5666c46e109119d1875fe3fc59f3b91bf6bf6e3/src/plugin.js#L315

davidswinegar:
@benmccann What do you think of the latest changes?

davidswinegar:
@benmccann I updated this to address your comments, how does it look now?

README.md
```diff
@@ -33,6 +33,11 @@ plugins: {
 
 			// Panning directions. Remove the appropriate direction to disable
 			// Eg. 'y' would only allow panning in the y direction
+			// A function that is called as the user is panning and returns the
+			// available directions can also be used:
+			//   mode: function({ chart }) {
+			//     return 'xy'
```
benmccann: I think you'd need a semicolon at the end of this line (same below)

src/plugin.js
```diff
@@ -73,6 +73,8 @@ function storeOriginalOptions(chart) {
 	});
 }
 
+// Mode can be a string ('x', 'y' or 'xy').
```
benmccann: How about making this jsoc syntax?

```
/**
 * @param {string} mode can be 'x', 'y' or 'xy'
 * @param {string} direction can be 'x' or 'y'
 */
```

benmccann:
lgtm. just two last minor comments. thanks for the patience on this

davidswinegar:
@benmccann updated to address your comments, thanks!

src/plugin.js
```diff
@@ -195,7 +199,7 @@ function doZoom(chart, percentZoomX, percentZoomY, focalPoint, whichAxes) {
 	if (zoomOptions.enabled) {
 		storeOriginalOptions(chart);
 		// Do the zoom here
-		var zoomMode = zoomOptions.mode;
+		var zoomMode = zoomOptions.mode === 'function' ? zoomOptions.mode({chart}) : zoomOptions.mode;
```
benmccann: Should it be `typeof zoomOptions.mode === 'function'` instead of `zoomOptions.mode === 'function'`? (same below)

README.md
```diff
@@ -69,6 +74,11 @@ plugins: {
 
 			// Zooming directions. Remove the appropriate direction to disable
 			// Eg. 'y' would only allow zooming in the y direction
+			// A function that is called as the user is zooming and returns the
+			// available directions can also be used:
+			//   mode: function({ chart }) {
+			//     return 'xy'
```
benmccann: there should be a semi-colon at the end of this line for consistency with above

src/plugin.js
```diff
@@ -195,7 +199,7 @@ function doZoom(chart, percentZoomX, percentZoomY, focalPoint, whichAxes) {
 	if (zoomOptions.enabled) {
 		storeOriginalOptions(chart);
 		// Do the zoom here
-		var zoomMode = zoomOptions.mode;
+		var zoomMode = typeof zoomOptions.mode === 'function' ? zoomOptions.mode({chart}) : zoomOptions.mode;
```
benmccann: this should be `{chart: chart}` to support older browsers

src/plugin.js
```diff
@@ -73,6 +73,10 @@ function storeOriginalOptions(chart) {
 	});
 }
 
+/**
+ * @param {string} mode can be 'x', 'y' or 'xy'
+ * @param {string} direction can be 'x' or 'y'
```
benmccann: direction should be dir to match the parameter name

PR #291, opened 2019-10-31
Fix #280 - Deprecation warning with Chart JS 2.9.1

### LINKED ISSUE #280: Deprecation warning with Chart JS 2.9.1 - time scale: "time.max" is deprecated. Please use "ticks.max" instead
When using this plugin (v 0.7.4), I see the following deprecation warning in the console when zooming my charts:

```Chart.js:1990 time scale: "time.max" is deprecated. Please use "ticks.max" instead```

devenbj:
Fix for Issue #280 by only using the time options if they are configured by the chart. This will enable backward compatibility while resolving the deprecation warnings for time axis charts with Chart JS 2.9.1.

Note: If the code calling the chart is still using time.min or time.max with Chart JS 2.9.1, it will continue to log deprecation warnings. This will ensure the chart still works.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -416,7 +416,7 @@ var zoomPlugin = {
 		chartInstance.$zoom = {
 			_originalOptions: {}
 		};
-		var node = chartInstance.$zoom._node = chartInstance.chart.ctx.canvas;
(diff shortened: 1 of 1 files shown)


devenbj:
Failed checks... will try again. :)

PR #292, opened 2019-10-31
Fix #280 - Deprecation warning with Chart JS 2.9.1

### LINKED ISSUE #280: Deprecation warning with Chart JS 2.9.1 - time scale: "time.max" is deprecated. Please use "ticks.max" instead

devenbj:
Fix for Issue #280 by only using the time options if they are configured by the chart. This will enable backward compatibility while resolving the deprecation warnings for time axis charts with Chart JS 2.9.1.

Note: If the code calling the chart is still using time.min or time.max with Chart JS 2.9.1, it will continue to log deprecation warnings. This will ensure the chart still works.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -169,8 +169,12 @@ function zoomTimeScale(scale, zoom, center, zoomOptions) {
 
 	var options = scale.options;
 	if (options.time) {
-		if (options.time.min) options.time.min = options.ticks.min;
-		if (options.time.max) options.time.max = options.ticks.max;
+		if (options.time.min) {
+			options.time.min = options.ticks.min;
+		}
+		if (options.time.max) {
+			options.time.max = options.ticks.max;
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


benmccann:
I think `min` and `max` are `undefined` by default, so I expect that this would cause them to never be set, which would mean zooming wouldn't work

devenbj:
With a time axis chart...
If you are using an older version of Chart.js or 2.9 (with deprecated `time.min` and `time.max`) the `min` and `max` values need to be set before the chart is drawn.... and thus would be set before panning or zooming could start. This plugin already updates `ticks.min` and `ticks.max`. This change tells the plugin to only clone the `ticks` options into the `time.min` and `time.max` options if the `time `options are in use on the current chart.

If you have updated your chart to use `ticks.min` and `ticks.max`, the plugin will still work because the `ticks.min` and `ticks.max` are set in the numerical functions (called by the time function).

Simply stated.... if the user isn't using `time.min` and `time.max`, we don't want them to be set by the plugin. If the user is using the `time` options, clone the values from `ticks.min` and `ticks.max`.

The change I created is a stop gap to cover users during transition from existing Chart JS versions until 3.0 (where the `time.min` and `time.max` options will no longer exist). It also eliminates thousands of "deprecated" `console.warn` entries when panning or zooming.

benmccann:
Ah yeah, I guess the scale sets `min` and `max` even if the user doesn't

PR #294, opened 2019-11-01
Unnecessary chart update?

HakuFlugel:
Not sure if I'm missing something, but this seems to cause an unnecessary amount of chart updates during mouse drag. Seems like removing it doesn't cause any issues down the line.

The performance gain is very noticable in big datasets.


HakuFlugel:
I was wondering how it drew the box when i dragged after removing it. I'm using the plugin in my vue app. For some unknown reason it draws the box even after removing the update.

I'll do some further investigation on this. Closing it for now.

Issue #319, opened 2020-02-05
drag-to-zoom does not work when `mode` is a function


SomeKittens:
This is because [`directionEnabled`](https://github.com/chartjs/chartjs-plugin-zoom/blob/6a97a22b99e0325597b7bd7b38b083dde4949a60/src/plugin.js#L80-L88) only handles the case where `mode` is `undefined` or a string.

Happy to submit a PR to fix this.

benmccann:
Sure, a PR would be welcome

PR #322, opened 2020-02-09
fix(mode): Fix bug where zoom did not work if mode was a function

SomeKittens:


src/plugin.js
```diff
@@ -76,12 +76,15 @@ function storeOriginalOptions(chart) {
...
+ * @param {Chart} chart instance of the chart in question
  */
-function directionEnabled(mode, dir) {
+function directionEnabled(mode, dir, chart) {
 	if (mode === undefined) {
 		return true;
 	} else if (typeof mode === 'string') {
 		return mode.indexOf(dir) !== -1;
+	} else if (typeof mode === 'function') {
+		return mode({ chart }).indexOf(dir) !== -1;
```
benmccann: Travis is failing. You'll have to change this to `{chart}`

SomeKittens:
Pushed a lint fix, and added tests for this case.  Confirmed that new tests fail on master without the fix in this PR.

SomeKittens:
Crosslinking: https://github.com/chartjs/chartjs-plugin-zoom/issues/319

SomeKittens:
Awesome, thanks for the quick merge!  Can you publish it to npm?

PR #338, opened 2020-03-20
fix zoom not working as intended in Firefox

### LINKED ISSUE #337: Zoom is not always working as intended in Firefox
The zoom does not work as intended in Firefox (Version: 74.0, Win 10, 64 bit). Firefox will sometimes fire the wheel event without the delta properties. This results in the zoom direction not being recognized correctly.

jelleschutter:
Zoom does not work as intended in Firefox since the wheel event will sometimes fire without the delta properties.

Closes #337

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -509,29 +509,31 @@ var zoomPlugin = {
 
 		var _scrollTimeout = null;
 		chartInstance.$zoom._wheelHandler = function(event) {
-			var rect = event.target.getBoundingClientRect();
-			var offsetX = event.clientX - rect.left;
-			var offsetY = event.clientY - rect.top;
+			if (typeof event.deltaY !== 'undefined') {
+				var rect = event.target.getBoundingClientRect();
+				var offsetX = event.clientX - rect.left;
+				var offsetY = event.clientY - rect.top;
 
-			var center = {
-				x: offsetX,
-				y: offsetY
-			};
+				var center = {
+					x: offsetX,
+					y: offsetY
+				};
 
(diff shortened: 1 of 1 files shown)


benmccann:
Does this fix it though? It looks like it would stop it from producing an error, but still wouldn't zoom like it should

Can you provide any more details about when the wheel event does or doesn't contain the delta properties? Is it always missing?

jelleschutter:
I've tested this on multiple sites and multiple devices. Firefox will always fire the wheel event twice: First without the delta and right after that once with the delta properties.

benmccann:
Thanks for the explanation. Can you add that in a comment and reverse the order of the `if` check? Here's what I mean:

```
// Prevent the event from triggering the default behavior (eg. Content scrolling).
if (event.cancelable) {
	event.preventDefault();
}

// Firefox always fires the wheel event twice:
// First without the delta and right after that once with the delta properties.
if (typeof event.deltaY === 'undefined') {
	return;
}

var rect = event.target.getBoundingClientRect();
.
.
.
```

src/plugin.js
```diff
@@ -509,6 +509,18 @@ var zoomPlugin = {
 
 		var _scrollTimeout = null;
 		chartInstance.$zoom._wheelHandler = function(event) {
+			// Prevent the event from triggering the default behavior (eg. Content scrolling).
+			if (event.cancelable) {
+				event.preventDefault();
+			}
+			
+
```
benmccann, line 517: thanks for making the changes. can you remove the duplicate blank line?

benmccann:
Thanks!

PR #341, opened 2020-04-13
feature: implement `threshold` for zoom operations

cnojima:
WHAT:
Implement a `threshold` option for Zoom.

WHY:
Certain scenarios trigger unwanted zoom operations - not necessarily a bug in the zoom code, but one that demands a bit more buffer before committing to a zoom.  I.e., if the chart is using CSS to fit a certain shape using percentages for width or height, a click (mousedown -> mouseup) could be misconstrued as a zoom operation.

HOW:
add an option, defaulting to the standard OOTB behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,9 @@ plugins: {
 			// On category scale, minimal zoom level before actually applying zoom
 			sensitivity: 3,
 
+			// Minimum mouse-movement delta before zoom is triggered, default `0`
+			threshold: 0,
+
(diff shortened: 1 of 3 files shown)


benmccann:
Someone else just submitted the same PR https://github.com/chartjs/chartjs-plugin-zoom/pull/347. I'm not sure why they submitted a new version, but I saw that one first as I was going down the list and merged it, so I guess I'll close this one since it's already been merged. Thanks for providing this!!

Issue #256, opened 2019-07-12
Zoom triggered when hiding/showing series by clicking legend


matcho:
Hi.

I have a chart with 2 or more series, drag zoom feature enabled, pan feature disabled, and chart legend enabled.

Sometimes when clicking multiple times the chart legend to show/hide one series or another, a zoom event is suddenly triggered (`onZoomComplete` callback is executed) and everything disappears: chart axis show ranges totally outside the data range.

Calling `resetZoom()` make things go back to normal.

I'm using the following configuration
```javascript
{
    zoom: {
        pan: {
            enabled: false,
            mode: "xy",
        },
        zoom: {
            enabled: true,
            drag: true,
            mode: "xy",
            onZoomComplete: someFunction
        }
    }
}
```

Thanks in advance for any help.

Mathias

benmccann:
What version are you using? I thought this was fixed in the latest version

matcho:
Thank you for your answer.
I'm using Chart.js version 2.7.3 and chartjs-plugin-zoom version 0.7.2, installed via NPM yesterday morning.

matcho:
Hi. I updated to 0.7.3 but the problem still happens. If I can send any more information or be useful in any way please let me know. All the best.

benmccann:
I'm not at my computer at the moment to test, but before it would happen with a click and that's what I thought was fixed. However, if you moved your mouse a very small distance while clicking that would be interpreted as a drag and zoom the chart. I wonder if you might be slightly moving the mouse without noticing. We perhaps should add a minimum distance the mouse must be moved to be considered a drag if that's the case.

matcho:
Oh yes you're right ! It's indeed the slight mouse movement that is triggering a drag event. Although the minimum distance trick sounds nice, perhaps it is possible to listen to mouse drag events on the "series" part of the chart only, and not the legend ?

benmccann:
Yeah. We should do both. I think this PR would fix it: https://github.com/chartjs/Chart.js/pull/6227

ikkala:
> I think this PR would fix it: chartjs/Chart.js#6227

I tried to apply this Chart.js 3.x fix to Chart.js 2.9.3 but it didn't seem to help in case of zoom plugin.
However, I did same kind of fix into zoom plugin directly -- I can make a pull request based on it.

ikkala:
[...]

Issue #317, opened 2020-01-29
Minimal step for "drag to zoom" on time chart


krulj:
I am trying to limit "sensitivity" of drag to zoom functionality, more precisely I want to start zooming only when user select at least 2% of chart. 

When user clicks on chart and start to drag (in order to zoom in), after releasing mouse button, I need to calculate if user selected just small part of chart (for example below 2% of whole chart):
- If yes, then chart will not be zoomed
- If no, then chart will be zoomed as usual

If someone done something similar, I appreciate your help!

cnojima:
I understand that this is an old issue.  I'm looking at the same need, i'm toying with a PR at the moment.

If you want a quick, hard-coded fix, change the `0` to `> 0` in this line: https://github.com/chartjs/chartjs-plugin-zoom/blob/master/src/plugin.js#L485

`2` works well to solve the issue i was trying to handle, other values may fit your use-case better.

If i do submit a PR, it'll be for a new option to set the drag-sensitivity to an integer.

benmccann:
Closed by https://github.com/chartjs/chartjs-plugin-zoom/pull/347

PR #352, opened 2020-05-07
Fix: Honor zoom.mode (xy/x/y) when applying zoom.threshold. Don't trigger zoom outside the area of chart lines/bars.

ikkala:
1. Enhances https://github.com/chartjs/chartjs-plugin-zoom/issues/317

I had case where options.zoom.mode was set to 'x' and threshold was set. The mouse dragging vertically was still triggering the zoom, even without exceeding the threshold.

2. Don't trigger zoom outside the area of chart lines/bars

If user starts mouse down for example in area of legends or axis the zoom starts. This fix limits the starting of zooming to the area of the actual graph.

Fixes https://github.com/chartjs/chartjs-plugin-zoom/issues/256

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "chartjs-plugin-zoom",
-  "version": "0.7.7",
+  "version": "0.7.8",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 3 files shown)


ikkala:
Sorry, didn't mean to include the commit 8c0b0d3 (0.7.8: Patched 0.7.7) here into pull request -- it was only for internal testing.

matcho:
Great news ! Thank you.

kurkle (commented):
This would collide with the `overScaleMode`, so would be best to add an option for the area limiting.

Needs a rebase.

Issue #115, opened 2017-12-07
Zoom ratio X:X 


joojoo102030:
I want to force the zoom ratio to always be x:x .
I am using bobble plots  which I need to be square  (x and y have the same scale and ticks )
I need the zoom to keep the ratio as well, or I could select a square inside the plot, and it zooms into that square 

Thanks..

PR #362, opened 2020-06-05
Keep aspect ratio on zoom with minmax range

aj8k:
I think that this issue is exactly about the problem I have faced with: https://github.com/chartjs/chartjs-plugin-zoom/issues/115

Charts that has min max ranges for zoom works with distortion when chart viewport could zoom out only into one direction, not in both directions due to range limitation. I have worked only with numeric charts. Maybe for other types such distortion is ok.


kurkle (commented):
This would need to be rebased. I think it would be best to put this functionality behind an option.

phimath:
Why was this closed? Was this ever merged?

PR #368, opened 2020-06-20
Add option pressed key CTRL for zoom

armagedon007:
Add option pressed key CTRL for zoom
if don't press CTRL key - stop zoom!


armagedon007:
Added an option for zoom when the CTRL key is pressed.

benmccann:
Closing in favor of https://github.com/chartjs/chartjs-plugin-zoom/pull/382. Better to make modifier key configurable than hardcode to Ctrl

PR #374, opened 2020-07-08
Expose changeZoom and changePan APIs

el:
We have a way to `resetZoom` but we dont have a way to change the zoom and pan. This PR exposes `doPan` and `doZoom` functions.

Sample usecase: https://codepen.io/elizzk/pen/OJMZMEm

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,14 @@ plugins: {
 
 ## API
 
+### chart.changePan(deltaX, deltaY)
+
+Programmatically change the pan. Pass differences for x and y axes.
+
+### chart.changeZoom(percentZoomX, percentZoomY)
+
+Programmatically change the zoom. Pass zoom percentages for x and/or y axes.
+
 ### chart.resetZoom()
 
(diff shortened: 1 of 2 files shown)


lcoenen-jci:
Hey that would be cool to be merged! I want to do something like this and without it I would have to use D3.

Koodos @el

el:
@benmccann Can this be merged ?

kurkle:
I think it would be best to export/expose those function in the plugin instead. We don't generally want to add to the chart instance.
It might require some refactoring.

el:
> I think it would be best to export/expose those function in the plugin instead. We don't generally want to add to the chart instance.
> It might require some refactoring.

I exposed them the same way resetZoom was exposed. If this is exposed as part of the plugin, the same should be the case for the resetZoom function.

kurkle:
> I exposed them the same way resetZoom was exposed. If this is exposed as part of the plugin, the same should be the case for the resetZoom function.

Exactly :)

PR #380, opened 2020-08-31
Add ctrlKeyWheelHorizontal and shiftKeyWheelVertical to zoom options.

paulcsf:
When `pan.enabled` is true and `zoom.drag` is false, the graph works nicely on mobile (pinch allows individual or both axis control), but there is no individual axis zoom control on desktop, only proportional zoom at the cursor location.

Enabling one or both of these options allows extending the desktop control for individual axis zoom.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,14 @@ plugins: {
 			//   },
 			mode: 'xy',
 
+			// Only zoom horizontally on ctrl+mouse-wheel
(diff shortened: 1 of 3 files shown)


benmccann:
Closing in favor of https://github.com/chartjs/chartjs-plugin-zoom/pull/382

Issue #379, opened 2020-08-31
Zoom is very slow in ChartJS Zoom Plugin-mode 'x' 


giaccomgl:
I'm trying to build a webpage with some graphs. In order to get it, I'm using ChartJS. Sometimes can happen there are too many data to see in just one canvas, so I need to use the ChartJS zoom plugin. I install and use it well, unless I have a big problem: Setting the mode as 'x', the zoom applied on canvas is really really slow (even if I put 100000 as speed value). I want to make it faster. How can I solve it? In this codepen you can see better what I mean:
https://codepen.io/giacomo-maraglino/pen/dyMRaRz

jepperaskdk:
Yep, same here. Hardly anything happens when trying to zoom.

leon9508:
hi, giaccomgl, I meet the same problem as you, have you solved it ?

leon9508:
Hi, I have read the source code just now,and know that how to make the zoom's speed of x-axis faster.
![image](https://user-images.githubusercontent.com/25746834/110352066-6ee31380-8070-11eb-924e-b84efff6f6d2.png)

PR #381, opened 2020-09-05
Zoom using zoomPercent

jepperaskdk:
Solves: https://github.com/chartjs/chartjs-plugin-zoom/issues/379

So, when using e.g. 1000 data-points, the zoom is slow in ticks of 1. It seems the indices were only changed in ticks of 1, rather than ticks of the percentage of the current-range.
This PR uses the speed property to set the percentage of the current range to be zoomed, e.g. with 1000 data points currently in view, a speed of 0.1 means we zoom 10% in on that range.

Next up should be to zoom at centerPointer, rather than either right or left.

### DIFF
--- a/samples/zoom-large-dataset.html
+++ b/samples/zoom-large-dataset.html
@@ -1,0 +1,85 @@
+<!doctype html>
+<html>
+
+<head>
+	<title>Bar Chart Zoom - Large dataset</title>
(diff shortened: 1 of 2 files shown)


ChristophKubik:
Same thing for panning would be nice as well

kurkle:
This was fixed by #434

Issue #377, opened 2020-08-13
Do not register plugin globally when using module import


qnp:
When using ES6 style modules (_e.g._ with a module bundler such as webpack), I would expect that when importing the plugin, it would **not** be registered globally to Chart, but instead the choice would be given to the programmer.

For instance, in my implementation with VueJS, I have a vue chart component wrapping Chart.js for which this plugin is activated conditionnaly by a prop. I have several dashboards for which this plugin is not used but doing such imports it is registered to Chart globally. Yet, this plugin is catching any touch events so that the dashboard cannot be scrolled down by a mobile user.

The actual workaround I found is to unregister the plugin right after the import:
```js
import chartJsPluginZoom from 'chartjs-plugin-zoom';
Chart.plugins.unregister(chartJsPluginZoom);
```
and later doing such a thing (simplified for the sake of the example):
```js
props: {
   zoomEnabled: {
      type: Boolean,
      default: false,
   },
},
mounted() {
  this.chart = new Chart(this.$refs.chartCanvas, {
     plugins: this.zoomEnabled ? [chartJsPluginZoom] : []
     ...
  })
}
```
So that the plugin is only registered in the Chart *instance* whenever I need it.

I guess that line should be removed:
https://github.com/chartjs/chartjs-plugin-zoom/blob/17b29b93ac27bfc36a77ed41b7dc263d1a574e01/src/plugin.js#L722

What do you think ?

Best regards

etimberg:
That appears to be the case as of April 25, 2021 in the master branch. https://github.com/chartjs/chartjs-plugin-zoom/blob/78f0386d33d5fd5ff4b46de92f03e4f83b0910a8/src/index.esm.js#L1-L3

qnp:
Awesome. Thanks for the feedback.

Issue #370, opened 2020-06-25
Activating zoom after first click


flaushi:
Is it possible to activate the zoom plugin only after the user clicks the canvas?

I have embedded several graphs in a long page, which has to be scrolled. It happens often that the cursor scrolls over the canvas (unintentionally), and then the next scroll event (mouse wheel) is eaten by the chart (instead of the main document).

I am afraid to destroy the mouse events if just intercepting 
```
$('#canvas').on('click', () => {
    chart.options.plugins.zoom.zoom.enabled = true; 
    chart.update();
});
```

Once zooming and panning is enabled, the click event is needed by the plugin, right?
Related issue: #335

cederlof:
#382 didn't do it for me, I landed on the code provided in OP's post. I also added `tabindex="0"` to the canvas so I could catch the `blur`-event for disabling zoom upon clicking outside of the chart.

Issue #335, opened 2020-03-15
Allow zooming only with a pressed modifier key


NicolasGoeddel:
Hi,

when I create a nearly page filling graph with the zoom functionality enabled, it is nearly impossible to find a spot where is no canvas to scroll the page. It would be nice if one could set a modifier key like Alt, Ctrl or Shift to enable zooming with the scroll wheel.

If there is already such a functionality, can you please give me an example application?

NicolasGoeddel:
I tried to change the `enabled` property with some key binds but although the property gets changed it does not enable the zooming functionality.

The `graphs` object stores all my graphs on the page and I iterate through them to change the state of the zoom.
```javascript
  $(document).keydown(function(event) {
    if (event.shiftKey) {
      for (var g in graphs) {
        graphs[g].options.plugins.zoom.zoom.enabled = true;
        console.log('enabled');
      }
    }
  });

  $(document).keyup(function(event) {
    if (event.shiftKey) {
      for (var g in graphs) {
        graphs[g].options.plugins.zoom.zoom.enabled = false;
        console.log('disabled');
      }
    }
  });
```

KoalaBear84:
What if you also add?

`graphs[g].update();`

etimberg:
A wheel modifier key was added in #382

Issue #308, opened 2019-12-26
Panning does not work with draggable-date plugin


Zuckerbrot:
I am using chart.js with the draggable data plugin
https://github.com/chrispahm/chartjs-plugin-dragdata

I would also like to use the panning functionality of the zoom-plugin.

However,I can only get either panning or data-dragging to work.
Example (works in Safari):
https://jsfiddle.net/zpq8an40/1/

This seems to be related to the fact that both panning and dragging use the same mouse gesture.

Would be nice to have both, like eg here
http://bl.ocks.org/stepheneb/1182434.

Would it be possible to allow the definition of a panKey, like {shift, ctrl, alt}, which, if pressed, toggles either the data-dragging or panning functionality?
This approach is used with highcharts to allow for both functionalities, cf. https://api.highcharts.com/highcharts/chart.panKey

Thanks very much and happy holidays!

benmccann:
I would be open to a pull request to add this functionality. I don't have time to add it myself

jledentu:
@Zuckerbrot Good idea. I'd suggest to call this new property `modifierKey`.

```js
plugins: {
	zoom: {
		pan: {
			enabled: true,
                        modifierKey: 'alt'
                },
                zoom: {
			enabled: true,
                        modifierKey: 'alt'
                }
        }
}
```

etimberg:
Looks like this got merged

PR #382, opened 2020-09-07
Added optional modifier keys

BarryPSmith:
Often the chart captures scroll events in an undesired way, so this PR makes it possible to set an optional `wheelModifierKey` that must be pressed to activate mouse wheel zoom.

A `modifierKey` can also be set for pan. By default this is only checked when the input comes via mouse, but can be set for touch and other inputs with `requireModifierNonMouse: true`

`zoom.onWheelModifierKeyFailed` is fired when a zoom does not occur because the modifier key is not present (e.g. to allow the consumer to display "Press ctrl to zoom"). A similar `pan.onModifierKeyFailed` is also available.

Although not necessarily the requested fix, it will help with the following issues:
#377, #370, #335, #308 (Actually I see the collaborators offer to accept PRs on this issue there). Not looking any further...

Relationship to other pull requests:
Might be incompatible with #380.
A more flexible version of #368

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ plugins: {
 			onPanComplete: function({chart}) { console.log(`I was panned!!!`); },
 			// Function called when pan fails because modifier key was not detected.
 			// event is the a hammer event that failed - see https://hammerjs.github.io/api#event-object
-			onModifierKeyFailed: function({chart, event}) { console.log(`I didn't start panning!`); }
+			onPanRejected: function({chart, event}) { console.log(`I didn't start panning!`); }
 		},
 
 		// Container for zoom options
@@ -119,7 +119,7 @@ plugins: {
 			// Function called once zooming is completed
 			onZoomComplete: function({chart}) { console.log(`I was zoomed!!!`); },
 			// Function called when wheel input occurs without modifier key
-			onWheelModifierFailed: function({chart, event}) { console.log(`I didn't start zooming!`); }
+			onZoomRejected: function({chart, event}) { console.log(`I didn't start zooming!`); }
 		}
 	}
 }
--- a/README.md
+++ b/README.md
@@ -62,7 +62,10 @@ plugins: {
 			// Function called while the user is panning
 			onPan: function({chart}) { console.log(`I'm panning!!!`); },
 			// Function called once panning is completed
-			onPanComplete: function({chart}) { console.log(`I was panned!!!`); }
+			onPanComplete: function({chart}) { console.log(`I was panned!!!`); },
+			// Function called when pan fails because modifier key was not detected.
+			// event is the a hammer event that failed - see https://hammerjs.github.io/api#event-object
+			onModifierKeyFailed: function({chart, event}) { console.log(`I didn't start panning!`); }
 		},
 
 		// Container for zoom options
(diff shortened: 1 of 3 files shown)


BarryPSmith:
@jledentu I see on issue #308 you suggest the name modifierKey for both the pan and zoom modifiers. I have used 'wheelModifierKey' for the zoom key in this PR because it only affects wheel zoom, but it would be an easy change to have them both 'modifierKey'.

NicolasGoeddel:
Cool. When can I expect the merge request to be accepted?

benmccann:
@BarryPSmith this will need to be rebased

benmccann:
Actually, thinking about this more, I think a better way to handle this might be a generic function that let's you filter whether the zoom occurs. That seems to handle more use cases. If I recall correctly that might already exist too in which case maybe we just need an example of using it for this usecase

NicolasGoeddel:
Okay, why the heck do you delete comments?

benmccann:
The comments are meant for code review. If you want to express interest you can simply leave a thumbs up on the PR

BarryPSmith:
It looks like enough has changed that it'll be easier to rewrite the PR. That's the impression I get looking at the conflict files that rebase has generated. At least it's a fairly simple set of changes to the functional code.

I'll throw away the zoom tests, given the modified test/utils.js has migrated into a different repo.

At least the weather is bad here so I have time.

kurkle:
some lint issues, `npm run lint` should reveal them lovally.

kurkle:
> Test coverage decrease: In order to test the new code one must be able to send mouse events with keys attached. That is not possible with the current test framework - util.js of chartjs-test-utils must be updated. See test/util.js of [d5cb18f](https://github.com/chartjs/chartjs-plugin-zoom/commit/d5cb18f90bb720e50cd930a047db9a5e6e8e0702)

Ok, I'll take a look at the tests later. I think it needs to be a WheelEvent for the delta stuff.

README.md
```diff
@@ -62,7 +62,10 @@ plugins: {
 			// Function called while the user is panning
 			onPan: function({chart}) { console.log(`I'm panning!!!`); },
 			// Function called once panning is completed
-			onPanComplete: function({chart}) { console.log(`I was panned!!!`); }
+			onPanComplete: function({chart}) { console.log(`I was panned!!!`); },
+			// Function called when pan fails because modifier key was not detected.
+			// event is the a hammer event that failed - see https://hammerjs.github.io/api#event-object
+			onModifierKeyFailed: function({chart, event}) { console.log(`I didn't start panning!`); }
```
kurkle, line 68: The event name is a bit surprising to me, how do you feel about `onPanRejectet`?

README.md
```diff
@@ -114,7 +117,9 @@ plugins: {
 			// Function called while the user is zooming
 			onZoom: function({chart}) { console.log(`I'm zooming!!!`); },
 			// Function called once zooming is completed
-			onZoomComplete: function({chart}) { console.log(`I was zoomed!!!`); }
+			onZoomComplete: function({chart}) { console.log(`I was zoomed!!!`); },
+			// Function called when wheel input occurs without modifier key
+			onWheelModifierFailed: function({chart, event}) { console.log(`I didn't start zooming!`); }
```
kurkle, line 122: Same here, how about `onZoomRejected`? it would be more generic and could be re-used for other use cases.

kurkle (commented):
Overall looks good. I'm not sure about the event names, is there some reasoning that I missed?

kurkle (approved):

kurkle:
Thanks @BarryPSmith

PR #414, opened 2020-12-30
Add an ESM build

joshkel:
Based off of chartjs-plugin-annotation's approach.

Specific changes:

- Switch from Gulp to Rollup for build; using Rollup directly works and seems simpler.
- Refactor some duplicate configuration in rollup.config.js.
- Change import style for Chart.js and Chart.js helpers so that it will work with modules.
- The new import style doesn't allow adding to the imported object, which caused problems for the `Chart.Zoom` namespace. Since that namespace appears to be undocumented and unused, I removed it.

The ESM version does not automatically register the plugin.  This matches chartjs-plugin-annotation and seems to better match Chart.js 3's tree-shaking design.

I'm unclear on whether the UMD build should automatically register the plugin. Not registering seems to better match Chart.js 3, and chartjs-plugin-datalabels [recently changed][1] to no longer register, but chartjs-plugin-annotation's UMD build does register.

[1]: <https://github.com/chartjs/chartjs-plugin-datalabels/commit/13daa478d9082fd0bc100e36394719215771d689>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,5 +9,3 @@ coverage/*
 
 nbproject/*
 dist/
-chartjs-plugin-zoom.js
-chartjs-plugin-zoom.min.js
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -33,17 +33,21 @@ gulp.task('test', gulp.parallel('lint', 'unittest'));
 gulp.task('watch', watchTask);
 gulp.task('default', gulp.parallel('lint', 'watch'));
 
+/*
+ * Create a zip file for distribution from the project's GitHub Releases page
+ */
 function packageTask() {
   return merge(
-    // gather "regular" files landing in the package root
+    // Gather "regular" files from dist. We don't pass a `base` prop, so gulp will
+    // put these in the top level of the zip archive.
     gulp.src([outDir + '*.js', 'LICENSE.md']),
 
-    // since we moved the dist files one folder up (package root), we need to rewrite
+    // Since we move the dist files one folder up in the archive, we need to rewrite
     // samples src="../dist/ to src="../ and then copy them in the /samples directory.
     gulp.src('./samples/**/*', {base: '.'})
       .pipe(streamify(replace(/src="((?:\.\.\/)+)dist\//g, 'src="$1')))
   )
-  // finally, create the zip archive
+  // Finally, create the zip archive.
   .pipe(zip(pkg.name + '.zip'))
   .pipe(gulp.dest(outDir));
 }
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -10,7 +10,6 @@ var inquirer = require('inquirer');
 var semver = require('semver');
 var path = require('path');
 var fs = require('fs');
-var {exec} = require('child_process');
 var karma = require('karma');
 var merge = require('merge-stream');
 var yargs = require('yargs');
@@ -24,17 +23,6 @@ var srcDir = './src/';
 var srcFiles = srcDir + '**.js';
 var outDir = './dist/';
 
-function run(bin, args, done) {
-  var exe = '"' + process.execPath + '"';
-  var src = require.resolve(bin);
-  var ps = exec([exe, src].concat(args || []).join(' '));
-
-  ps.stdout.pipe(process.stdout);
-  ps.stderr.pipe(process.stderr);
-  ps.on('close', () => done());
-}
-
-gulp.task('build', gulp.series(rollupTask, copyDistFilesTask));
 gulp.task('package', packageTask);
 gulp.task('bump', bumpTask);
 gulp.task('lint-html', lintHtmlTask);
@@ -43,20 +31,7 @@ gulp.task('lint', gulp.parallel('lint-html', 'lint-js'));
 gulp.task('unittest', unittestTask);
 gulp.task('test', gulp.parallel('lint', 'unittest'));
 gulp.task('watch', watchTask);
-gulp.task('default', gulp.parallel('lint', 'build', 'watch'));
-
-function rollupTask(done) {
-  run('rollup/dist/bin/rollup', ['-c'], done);
-}
-
-/**
- * Copy the files from `/dist` to the root directory.
- * @todo remove at version 1.0
- */
-function copyDistFilesTask() {
-  return gulp.src(outDir + '*.js')
-    .pipe(gulp.dest('./'));
-}
+gulp.task('default', gulp.parallel('lint', 'watch'));
 
 function packageTask() {
   return merge(
@@ -151,5 +126,5 @@ function unittestTask(done) {
 }
 
 function watchTask() {
-  return gulp.watch(srcFiles, gulp.parallel('lint', 'build'));
+  return gulp.watch(srcFiles, gulp.parallel('lint'));
 }
--- a/package.json
+++ b/package.json
@@ -6,12 +6,13 @@
   "jsdelivr": "dist/chartjs-plugin-zoom.min.js",
   "unpkg": "dist/chartjs-plugin-zoom.min.js",
   "main": "dist/chartjs-plugin-zoom.js",
+  "module": "dist/chartjs-plugin-zoom.esm.js",
   "repository": {
     "type": "git",
     "url": "https://github.com/chartjs/chartjs-plugin-zoom.git"
   },
   "scripts": {
-    "build": "gulp build",
+    "build": "rollup -c",
     "test": "gulp test"
   },
   "devDependencies": {
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -183,7 +183,7 @@ function doZoom(chart, percentZoomX, percentZoomY, focalPoint, whichAxes, animat
 		// Do the zoom here
 		var zoomMode = typeof zoomOptions.mode === 'function' ? zoomOptions.mode({chart: chart}) : zoomOptions.mode;
 
-		// Which axe should be modified when figers were used.
+		// Which axes should be modified when fingers were used.
 		var _whichAxes;
 		if (zoomMode === 'xy' && whichAxes !== undefined) {
 			// based on fingers positions
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -1,12 +1,11 @@
 'use strict';
 
-import Chart from 'chart.js';
+import * as Chart from 'chart.js';
+import * as helpers from 'chart.js/helpers';
 import Hammer from 'hammerjs';
 
-var helpers = Chart.helpers;
-
-// Take the zoom namespace of Chart
-var zoomNS = Chart.Zoom = Chart.Zoom || {};
+// Zoom namespace (kept under Chart prior to Chart.js 3)
+var zoomNS = {};
 
 // Where we store functions to handle different scale types
 var zoomFunctions = zoomNS.zoomFunctions = zoomNS.zoomFunctions || {};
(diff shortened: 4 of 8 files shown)


package.json
```diff
@@ -6,12 +6,13 @@
   "jsdelivr": "dist/chartjs-plugin-zoom.min.js",
   "unpkg": "dist/chartjs-plugin-zoom.min.js",
   "main": "dist/chartjs-plugin-zoom.js",
+  "module": "dist/chartjs-plugin-zoom.esm.js",
   "repository": {
     "type": "git",
     "url": "https://github.com/chartjs/chartjs-plugin-zoom.git"
   },
   "scripts": {
-    "build": "gulp build",
```
benmccann, line 14: can you remove the `build` command from the gulpfile now that it's unused?

benmccann:
Awesome! Thanks for this!

It looks like `chartjs-plugin-datalabels` was changed to not register so that the plugin isn't automatically enabled on all charts. I agree you may want to enable a plugin only on some charts. I also agree we should probably make sure this plugin, the data labels plugin, and annotations plugin work the same way. @etimberg @kurkle do you have thoughts on UMD vs ESM behavior here and what the standard should be?

kurkle:
I think we agreed on not registering on any build to be consistent.

package.json
joshkel: @benmccann Fixed. Thanks.

joshkel:
After I asked yesterday about whether plugins should register, I remembered that the v3 migration guide [says](https://www.chartjs.org/docs/master/getting-started/v3-migration/#setup-and-installation) that Chart.js 3 itself doesn't require registering when it's imported via a `<script>` tag (which, if I understand correctly, would be when using the UMD build).

So registering in the UMD build seems more consistent?

(I'm happy to do whichever; I just wanted to confirm before proceeding.)

simonbrunel:
The rationale behind this change in the datalabels plugin is detailed in chartjs/chartjs-plugin-datalabels#42.

> So registering in the UMD build seems more consistent?

Auto-registering should have nothing to do with the type of package, especially because UMD builds can be used with `require` and `import` (not only `<script>`). So it's quite inconsistent that the same import syntax against different packages doesn't produce the same result.

And when using `<script>`, users still have to write JavaScript in order to create a chart, so I don't see any benefit of auto-registering plugins in this case. It brings inconsistencies between import methods, makes more complicated the build/release process and needs to be explained in the docs. All of this to prevent `<script>` users to write `Chart.register(plugin)`.

kurkle:
Its not only about plugins though. If we want to be consistent  the registration of controllers, elements and scales should be required in all builds too. and that would be silly in my opinion, because you can not shake anything from an umd build. unless we provide some generic `registerAll`. And if we do that, maybe plugins should do the same?

simonbrunel:
It's not the same because registering a plugin makes it immediately active while registering a controller / element / scale does not. I thinks I would actually make registering core plugins required for all builds so it's fully consistent. And again, registering a plugin is not about tree-shaking but about making it "active" for all charts. The fact that there is a single `Chart.register()` method for all the things may make things a bit confusing.

simonbrunel:
> unless we provide some generic `registerAll`

I find **extremely** annoying, not intuitive and not user friendly the way to integrate Chart.js with the ESM build (which is nowadays the most common use case). I understand why "some" wants this granularity of import/register but "most" users should not have to deal with [all these imports](https://www.chartjs.org/docs/master/getting-started/integration#bundlers-webpack-rollup-etc). For example, a user who wants to create a line chart should not have to think about registering the `LineElement`, `PointElement`, `LineController`, `CategoryScale`, `LinearScale`, etc. because these are implicit dependencies coming from defaults. 

One solution could be to provide some kind of "presets" so "most" users would just have to:

```js
import { LineChart } from 'chart.js';

Chart.register(LineChart);

new Chart(ctx, {
    type: 'line',
    options,
    data,
});
```

benmccann:
There are some such as pie where registration is probably unnecessarily verbose even for tree-shaking users - you can't have a pie / doughnut / polar area chart without radial scale and arc element, so perhaps we should automatically register those for you in that case. 

I kind of like the preset idea. LineController is one of the trickiest because it's pretty common not to use the default scales. E.g. the category scale is often switched out for linear or time/timeseries scale.  Line/radar is also the only one with multiple possible elements. Registering both elements would make sense for the preset case. Would we register all compatible scales as well? And also all compatible plugins like the filler plugin?

benmccann:
I think the idea of registering and enabling plugins has gotten a bit muddied. I think plugins should always be registered when imported (i.e. available for use), but should not be globally enabled by default in either the UMD or ESM builds. I think it's reasonable to have an API to globally enable, but we should not call it `register` as it's an overloaded use of the term and we should call it `enable` or something else.

simonbrunel:
> I think plugins should always be registered when imported (i.e. available for use)

I disagree on this because it's usually not a good practice to expect side effect on import. Also, we could imagine a plugin that "overrides" another plugin, in which case we don't necessarily want the base plugin to be registered. For example:

```js
import DataLabels from 'chartjs-plugin-datalabels';

// chartjs-plugin-custom-datalabels
export default Object.assign({}, DataLabels , {
   beforeUpdate() {
      // ...
   }
});
```

I would compare this to Vue.js and their official plugins: they recently chose to remove this side-effect logic in favor of explicit registration. I also think it's a better approach, at least for our plugins. And honestly, I'm not sure to understand what we are trying to simplify?

```js
import ZoomPlugin from 'chartjs-plugin-zoom';

Chart.plugins.register(ZoomPlugin);
```

vs


```js
import 'chartjs-plugin-zoom';  // side-effect

Chart.plugins.enable('zoom');
```

The first snippet looks more predictable and robust IMO.

simonbrunel:
> I think plugins should always be registered when imported (i.e. available for use)

@benmccann after reading again your comment, I'm wondering if I correctly understood your suggestion :) i.e. registering the plugin should be a side effect of importing the plugin or if we should expect the user to explicitly call `Chart.register()` after the import (i.e. the change I made for the datalabels plugin)?

benmccann:
I was suggesting to have a side effect, but to your point that's not really necessary to avoid verbosity. Maybe a better way to solve my concern without side effects is to rename the API. `register` to me generally just means to make something available in the registry and I didn't like that it's also having another effect of also globally enabling a plugin. Something like `registerAndEnable` might be more explicit.

benmccann:
@joshkel could you rebase this since https://github.com/chartjs/chartjs-plugin-zoom/pull/416 was merged?

The question about whether to register in the UMD builds seems to be a big one :smile: Perhaps just leave it out for now and we can discuss the best path forward separately without blocking your PR

jledentu:
I agree with @simonbrunel that the first syntax is much more explicit, and it's a good practice to avoid side effects IMO.

@benmccann Maybe a function named `use` (like in Vue) would be even more explicit (we don't only "register" something but really "use" it):

```js
Chart.plugins.use(ZoomPlugin);
```

src/plugin.js
```diff
@@ -1,12 +1,10 @@
 'use strict';
 
-import Chart from 'chart.js';
+import * as helpers from 'chart.js/helpers';
```
kurkle, line 3: only `each` and `isNullOrUndef` is used, so maybe just import those?

kurkle:
> @joshkel could you rebase this since #416 was merged?
> 
> The question about whether to register in the UMD builds seems to be a big one 😄 Perhaps just leave it out for now and we can discuss the best path forward separately without blocking your PR

I agree to this, the removal of auto-registration needs changes in test setup, so it should be done separately.

joshkel:
Thanks for the discussion and feedback. I've rebased and responded to the remaining code review comments.

Switching to an explicit use / register step makes sense to me, but as suggested by @kurkle, I preserved the existing behavior (UMD _does_ auto-register) for now, to keep this PR more focused and to leave the tests unchanged.

benmccann (approved):

gulpfile.js
```diff
@@ -43,20 +31,7 @@ gulp.task('lint', gulp.parallel('lint-html', 'lint-js'));
...
-
-function rollupTask(done) {
-  run('rollup/dist/bin/rollup', ['-c'], done);
-}
-
-/**
- * Copy the files from `/dist` to the root directory.
- * @todo remove at version 1.0
- */
-function copyDistFilesTask() {
```
benmccann, line 56: removing this will break `gulp package` so that the release process doesn't work. you might want to leave this but make it run as the first step of `package` instead of the last step of `build`

.gitignore
```diff
@@ -9,5 +9,3 @@ coverage/*
 
 nbproject/*
 dist/
-chartjs-plugin-zoom.js
```
benmccann, line 12: you should probably leave these and add the esm files as well for packaging. see my other comment below

benmccann (changes requested):

.gitignore
joshkel: If I'm understanding correctly, we don't need these .gitignore lines or the gulpfile `copyDistFilesTask`.

The previous configuration invoked Rollup to put the files under `dist`, then it called `copyDistFilesTask` to copy the files out of `dist` into the project root (which is why the files are listed in .gitignore here). But package.json references the copies under `dist` directory, and the gulpfile's `packageTask` references the copies under `dist` (via `outDir`). I can't tell that the copies in the project root are used at all, so I removed them.

If I've missed something, please let me know.

benmccann: `gulp package` looks like it looks for files in the root directory: https://github.com/chartjs/chartjs-plugin-zoom/blob/8194ad68db5d197c7eea539fc2a4b4bd4ef8c082/gulpfile.js#L64

Alternatively, you might be able to update `gulp package` to look in `dist`

joshkel: If I'm understanding Gulp correctly, it treats `outDir` as the [glob base](https://gulpjs.com/docs/en/api/concepts#glob-base) for `outDir + '*.js'`, so the files from `dist` are included at the top level of the resulting zip file. Contrast with the explicit `base` parameter for the `samples` glob; those are included in a `samples` subdirectory within the zip file:

https://github.com/chartjs/chartjs-plugin-zoom/blob/8194ad68db5d197c7eea539fc2a4b4bd4ef8c082/gulpfile.js#L68

I tested by running `npm run build; npx gulp package` in a fresh copy of my branch, and it created a `dist/chartjs-plugin-zoom.zip` file with contents similar to the releases on GitHub, so I think things are working.

kurkle (approved):

.gitignore
benmccann: Ah, we might need to update the comments in the `packageTask` as they threw me off and look like they might be outdated:
> // gather "regular" files landing in the package root
> // since we moved the dist files one folder up (package root), we need to rewrite
> // samples src="../dist/ to src="../ and then copy them in the /samples directory.

joshkel: Yes, I found that confusing too. I just pushed some comments to try and clarify; feedback welcome.

kurkle (approved):
A lot clearer to me

benmccann:
Awesome. Thanks for this!!

If you have any interest in contributing further, I'd love to get rid of `gulp` altogether :smile:

joshkel:
Sure, I can take a look at removing gulp. Thank you.

benmccann:
@joshkel it's gone now :smile: @kurkle was very fast as making a whole bunch of improvements! thanks again for the help

kurkle:
> Sure, I can take a look at removing gulp. Thank you.

Oh, sorry! I did not see this 😞

joshkel:
No problem; I hadn't started yet. Thanks, @kurkle and @benmccann.

PR #415, opened 2020-12-30
exported doZoom function

laithyounes:
this could help to use the doZoom function and trigger it using other methods

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -672,4 +672,6 @@ var zoomPlugin = {
 };
 
 Chart.register(zoomPlugin);
+
+export { doZoom };
 export default zoomPlugin;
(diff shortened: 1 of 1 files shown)


benmccann:
Can you explain the use case? Usually you can just manually set `min` and `max` on the scale and don't really need to call `doZoom` if you're triggering this in another way

Also, this is failing the CI

laithyounes:
> Can you explain the use case? Usually you can just manually set `min` and `max` on the scale and don't really need to call `doZoom` if you're triggering this in another way
> 
> Also, this is failing the CI

Hey @benmccann, I want to create a zoom in/out buttons beside the current functionality,
using the doZoom function will provide this functionality without effecting the animation

PR #418, opened 2021-01-05
Restore .travis.yml

benmccann:
I setup GitHub Actions in https://github.com/chartjs/chartjs-plugin-zoom/pull/402, but was too hasty in removing the `.travis.yml` file. We still need it for releases


kurkle:
I did remove the api hook too yesterday. So need to restore that too.

benmccann:
what's the api hook?

PR #419, opened 2021-01-05
Add release-drafter and publish-npm workflows

kurkle:


benmccann:
Can you update https://github.com/chartjs/chartjs-plugin-zoom/blob/master/MAINTAINING.md with how this works? I'd love to get rid of Travis altogether :smile: I think it might be broken...

kurkle:
This just drafts the release notes for github release. I'll create a PR for using actions for releasing and update the MAINTAINING.md in that.

Issue #426, opened 2021-02-23
Add SRI to script tags in README.md


jacopotediosi:
Hi everyone!
IMHO security is important and sometimes an understimate topic.

I noticed that in README.md script tags don't include SRI hashes (https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity), so maybe many web devs who are in a hurry are copying and pasting them missing this basic security feature (I was going to do it too).

Since the effort to fix in this case is really minimal (simpy add SRI to script tags in README.md file), I open this issue to ask if my question could be compatible with the purposes of this project and to hear what the community thinks about that.

I know that the chartjs documentation itself also suffers from this problem (e.g. https://www.chartjs.org/docs/latest/getting-started/), but for that I will open a separate issue on the correct repository.

PR #427, opened 2021-02-23
Replace script tags with script+SRI

jacopotediosi:
Please see issue #426

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,9 +16,9 @@ Run `npm install --save chartjs-plugin-zoom` to install with `npm`.
 If including via a `<script>` tag, make sure to include `Hammer.js` as well:
 
 ```html
-<script src="https://cdn.jsdelivr.net/npm/chart.js@2.9.3"></script>
-<script src="https://cdn.jsdelivr.net/npm/hammerjs@2.0.8"></script>
-<script src="https://cdn.jsdelivr.net/npm/chartjs-plugin-zoom@0.7.7"></script>
+<script src="https://cdn.jsdelivr.net/npm/chart.js@2.9.3/dist/Chart.min.js" integrity="sha256-R4pqcOYV8lt7snxMQO/HSbVCFRPMdrhAFMH+vr9giYI=" crossorigin="anonymous"></script>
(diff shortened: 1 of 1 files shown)


benmccann:
Is it possible to add the `integrity` attribute without adding `/dist/Chart.min.js` in the `src`?

jacopotediosi:
> Is it possible to add the `integrity` attribute without adding `/dist/Chart.min.js` in the `src`?

Of course, but please note that if the CDN maintainer changes the default file (for example, it sets the .js file instead of the .min.js one) the integrity will no longer be valid (because the hash was calculated for the original file) and all the sites where the script was included will be broken (such an action would still be very impactful and risky even without SRI, and this is why in all my sites I specify the exact file I intend to include).

benmccann:
That seems unlikely to be much of a risk since we're using a versioned package

kurkle:
I think it would be better to write some instructions, that do not include maintenance overhead here, when updating versions.
Its bad practice to copy the SRI from anywhere else than the CDN you are relying on.

jacopotediosi:
> I think it would be better to write some instructions, that do not include maintenance overhead here, when updating versions.
> Its bad practice to copy the SRI from anywhere else than the CDN you are relying on.

Clicking "Copy HTML" button on jsdeliver.net writes the complete SRI hash tag to the clipboard, so I don't see too much overhead. Please note that if the SRI hash doesn't match the file pointed by src url (eg you copied it from a bad site which is trying to do something malicious) the browser will refuse to load the script. So the best practice is to check the correctness of the src; there are no security problems if you copy the SRI from places other than the CDN.

kurkle:
Yeah, its easy to get the SRI from the CDN, once the package is published and available from the CDN.
The version needs to be updated when a new release is published. Ideally everything is updated before the new version is published. Either calculating the SRI from build or updating the README after release, in my opinion counts for maintenance overhead. 

If we decided to add the SRI to script tags in docs, it should of course be applied to all repositories under chartjs.

etimberg:
If we can find a way to autogenerate this, that'd be the best solution in my opinion. I agree with @kurkle that this adds some maintenance burden for new releases that can only be done after the code is tagged, so the changes won't be in the actual release tag.

kurkle:
Closing. If you find an way to autogenerate those when releasing to NPM, happy to take another look.

PR #439, opened 2021-03-11
Reuse types from Chart.ts

tomsoftware:
Reuse types from Chart.js ??
->This might be a good idea cuase chartjs-plugin-zoom is strongly bundled with chartjs?

### DIFF
--- a/types/options.d.ts
+++ b/types/options.d.ts
@@ -1,19 +1,24 @@
-import { Chart, Point, Color } from 'chart.js';
+import { Chart, Color } from 'chart.js';
 
 
 type Mode =  'x' | 'y' | 'xy';
 
-interface DragEffectOptions {
+export interface DragEffectOptions {
 	borderColor?: Color;
 	borderWidth?: number;
 	backgroundColor?: Color;
 	animationDuration?: number;
 }
 
+export interface RangePoint {
+	x?: number | string;
+	y?: number | string;
+}
+
 /**
  * Container for zoop options
  */
-interface ZoomOptions {
+export interface ZoomOptions {
 	/**
 	 * Boolean to enable zooming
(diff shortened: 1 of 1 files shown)


types/options.d.ts
```diff
@@ -27,26 +25,29 @@ interface ZoomOptions {
...
+	 */	
 	mode: Mode | { (char: Chart): Mode };
 
 	/**
-	  * Format of min zoom range depends on scale type
-	  */
-	rangeMin?: XyValue;
+	 * Format of min zoom range depends on scale type
+	 */
+	rangeMin?: Point;
```
kurkle, line 42: I think `number | string` describes these better. Point is `{x, y}`, which is not correct for these.

(So the `XyValue` is also incorrect)

tomsoftware: Sorry, I used the "README.md" as source

```
rangeMin: {
    // Format of min zoom range depends on scale type
    x: null,
    y: null
   },
```

kurkle: I stand corrected, proper type would be `{x: number | string, y: number | string}`

kurkle (approved):

types/options.d.ts
tomsoftware: ok, I'm not surce if
```
 const rangeMin = zoomPanOptions.rangeMin[zoomPanOptions.scaleAxes];
    if (newMin < rangeMin) {
```
is meant for a string comparison... but doesn't matter :-)

Thanks for so mutch for your work and hopefully this will be a good starting point for you and others :-)

kurkle: Right, I though it was available for all scales. Seems it only works for numeric scales. It'll get fixed at some point in some direction :)

PR #451, opened 2021-04-14
Remove animationDuration option

### LINKED ISSUE #450: `animationDuration` doesn't correctly define the duration of drag animation
Hi, :wave: 

In the current master version, the `animationDuration` option doesn't work correctly. It only allows to enable/disable the animation on drag, but the duration is not taken into account.

In the code, the duration is set on `chart.options.animation.zoom.duration`, but it seems that we should set `chart.options.transitions.zoom.animation.duration` instead (I guess but I'm not sure to understand how transitions work in Chart.js 3).

By the way, I don't know if the `animationDuration` option is still needed since we can know customize the `zoom` transition in `chart.options.animation.zoom`?

* chart.js 3.0.1
* chartjs-plugin-zoom master

jledentu:
As explained in #450, I suggest to remove the `animationDuration` option (not working with Chart.js 3) in favor of configuration of the `zoom` transition.

Closes #450

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -184,9 +184,9 @@ function zoomScale(scale, zoom, center, zoomOptions) {
  * @param {number} percentZoomY The zoom percentage in the y direction
  * @param {{x: number, y: number}} [focalPoint] The x and y coordinates of zoom focal point. The point which doesn't change while zooming. E.g. the location of the mouse cursor when "drag: false"
  * @param {string} [whichAxes] `xy`, 'x', or 'y'
- * @param {string} [transitionMode] Mode used for update
+ * @param {boolean} [useTransition] Whether to use `zoom` transition
  */
-function doZoom(chart, percentZoomX, percentZoomY, focalPoint, whichAxes, transitionMode) {
+function doZoom(chart, percentZoomX, percentZoomY, focalPoint, whichAxes, useTransition) {
   var ca = chart.chartArea;
   if (!focalPoint) {
     focalPoint = {
@@ -225,11 +225,7 @@ function doZoom(chart, percentZoomX, percentZoomY, focalPoint, whichAxes, transi
       }
     });
 
-    if (transitionMode && chart.options.transitions[transitionMode]) {
-      chart.update(transitionMode);
-    } else {
-      chart.update('none');
-    }
+    chart.update(useTransition ? 'zoom' : 'none');
 
     if (typeof zoomOptions.onZoom === 'function') {
       zoomOptions.onZoom({chart: chart});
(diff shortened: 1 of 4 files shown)


src/plugin.js
```diff
@@ -225,15 +225,8 @@ function doZoom(chart, percentZoomX, percentZoomY, focalPoint, whichAxes, animat
...
-    if (animationDuration) {
-      // needs to create specific animation mode
-      if (!chart.options.animation.zoom) {
-        chart.options.animation.zoom = {
-          duration: animationDuration,
-          easing: 'easeOutQuad',
-        };
-      }
-      chart.update('zoom');
+    if (transitionMode && chart.options.transitions[transitionMode]) {
```
kurkle, line 228: I think it could be just `chart.update('zoom')`. Not sure where we'd need to use another transition mode for it?

src/plugin.js
```diff
@@ -650,7 +643,7 @@ var zoomPlugin = {
           delete scaleOptions.max;
         }
       });
-      chartInstance.update();
+      chartInstance.update('none');
```
kurkle, line 646: Why `none` mode here? I think it would probably be ok to use `zoom` here as well.

kurkle (commented):
Looks like good direction to me!

src/plugin.js
jledentu: Yeah, I had left the door open to more customization in other zoom modes, but maybe unnecessary. I can just leave a `useTransition` parameter in the `doZoom` function:

```js
if (useTransition) {
  chart.update('zoom');
}
```

?

src/plugin.js
kurkle: That might be a good idea. Lets leave it out from this pr still though and just use the `none` for now, because that probably meets most use cases.

src/plugin.js
kurkle: I don't think we need to check if the transition is configured, it will fall back to default animations if not. And we need to update in either case. Also feels like the `useTransition` is not needed, because one can configure the zoom transition to have no duration.

jledentu: Ok I agree for the `chart.options.transitions.zoom` check. But we need a `useTransition` parameter because there are some calls to `doZoom` where we don't want any animation. For instance when zooming with mouse scroll, using an animation creates a "latency" which is very annoying.

jledentu: I updated the PR

kurkle: Good point, sheds some light to your thinking of possibly needing separate transitions too.

It could be ideal if we could provide some additional info in the context of the transition:

```js
options: {
  transitions: {
    zoom: {
      animation: {
        duration: ctx => ctx.mode === 'scroll' || ctx.mode === 'reset' : 0 ? 500
      }
    }
  }
}
```

But as we can't we need to separate by transition.
So eventually we might have `zoomScroll`, `zoomDrag`, `zoomReset`, `pan` transitions?

Back to this PR, I think `useTransition` is good enough for this bug fix.

kurkle (approved):

Issue #441, opened 2021-03-15
resetzoom triggering after drag (beta)


pgp80:
I'm trying out the betas of both chartjs and zoom, and I'm noticing some behavior which is different from the current versions I have on a project. I have drag zoom enabled and in the current version, it zooms and then I have onClick: resetZoom and it reverts to original form. In the beta, I drag and as soon as I release the mouse the resetZoom occurs immediately (so it doesn't stay zoomed in).  Removing the onClick in the beta does result in a stable zooming. Just thought I'd share!
``  
options: {
  
    
    scales: {
      x: {
      	type: 'time',
          unit: 'hour'
      },//end x
      y: {
        
      }//end y
    },//end scales
     
    plugins: {
      zoom: {
        
        zoom: {
          enabled: true,
          drag: true,
          mode: 'x',
        
        }
      }
    },
    onClick: resetZoom,
 }//end options

![recording (1)](https://user-images.githubusercontent.com/60016497/111176710-f3a5c400-857f-11eb-936a-4881758a1c11.gif)

kurkle:
Thanks for the heads up, I'm guessing it should not call "onClick" when dragging.

jledentu:
@kurkle  I think it's the native behavior to have `mouseup` and `click` events triggered simultaneously, even if `mousemove` events were triggered between the `mousedown` and `mouseup`. I don't know why it worked in previous versions, but I think that a `onClick` callback is expected when a `click` event natively happens, don't you think?

@pgp80 There are some tricks to distinguish quick click and "drag" events: https://stackoverflow.com/questions/6042202/how-to-distinguish-mouse-click-and-drag.

pgp80:
I've been trying the latest betas and the drag still appears to be triggering the onClick: resetZoom on my end. I suppose I can try the tricks per the above, but just figured I'd let you know in case this still isn't intended behavior, thanks!

kurkle:
@pgp80 is your click handler on the chart options or in the canvas element?

PR #455, opened 2021-04-16
Refactor startHammer

kurkle:
* split `startHammer` into three parts
* remove the `_ghostClickHandler` that would mask the `click` event on pan end of the canvas
  * this must have somehow been inserted before chart event listeners in v2
  * add similar logic in `beforeEvent` to mask the `click`

Resolves: #441

### DIFF
--- a/src/hammer.js
+++ b/src/hammer.js
@@ -63,33 +63,29 @@ function createPinchHandlers(chart, zoomOptions) {
 }
 
 function createPanHandlers(chart, panOptions) {
-  let prevDeltaX = null;
-  let prevDeltaY = null;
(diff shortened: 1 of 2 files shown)


Koxx3:
hello,
sorry, but I still struggle with scrolling pages with chartjs & zoom plugin.
can you explain the important options to enable page scrolling (and allow pinch zoom) please ?

marttheis:
Hello,

I am also still having the problem that scroll gestures are eaten by chartjs when done on the chart. I am using 
Chartjs 4.3.0
zoom plugin 2.0.1
hammerjs 2.08

It doesn't work.

Issue #444, opened 2021-04-06
Programmatically zoom


KoalaBear84:
Is there a way to zoom programmatically? Like "Zoom to last day"?

I ask this for the Chart.js 3 version.

stockiNail:
@KoalaBear84 maybe you should create an event and dispatch it.

When the plugin is enabled, the element used to get events by HammerJS is stored in `chartInstance.$zoom._node`.

```javascript
onMyZoom = function() {
  if (chartInstance.$zoom && chartInstance.$zoom._node) {
    const wheeleventLeft = new WheelEvent("wheel", { deltaX: -500 });
    chartInstance.$zoom._node.dispatchEvent(wheeleventLeft);
  }
}
```

KoalaBear84:
Thanks @stockiNail for your quick answer, and sorry for the late reply.

```javascript
yourchart.config.options.plugins.zoom.zoom.rangeMin.x = new Date('2021-04-19');
yourchart.config.options.plugins.zoom.zoom.rangeMax.x = new Date('2021-04-20');
yourchart.$zoom._node.dispatchEvent(new WheelEvent("wheel", { deltaX: 0, deltaY: 0, delta: 0 }))
```

This sort of works and was what I hoped be possible. The problem is that this is now a 'hard' limit and have to use `resetZoom` to 'exit'. Is there any other way? Also, when calling the event, it always moves, even though all delta's are 0.

Do you have idea's for these?

etimberg:
Duplicate of #82

PR #464, opened 2021-04-25
Refactor & export pan and zoom functions

kurkle:
* Simplify the `doPan` and `doZoom` interface.
* export doPan, doZoom and resetZoom functions.
* add `pan` and `zoom` functions to chart instance.
* Add sample
*

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -43,14 +43,14 @@ export function doZoom(chart, zoom, options = {}, useTransition) {
 
   storeOriginalScaleLimits(chart);
 
-  const xEnalbed = directionEnabled(mode, 'x', chart);
-  const yEnabled = directionEnabled(mode, 'y', chart);
+  const xEnalbed = x !== 1 && directionEnabled(mode, 'x', chart);
+  const yEnabled = y !== 1 && directionEnabled(mode, 'y', chart);
   const enabledScales = overScaleMode && getEnabledScalesByPoint(overScaleMode, focalPoint, chart);
 
   each(enabledScales || chart.scales, function(scale) {
-    if (x !== 1 && scale.isHorizontal() && xEnalbed) {
+    if (scale.isHorizontal() && xEnalbed) {
       zoomScale(scale, x, focalPoint, options);
-    } else if (y !== 1 && !scale.isHorizontal() && yEnabled) {
+    } else if (!scale.isHorizontal() && yEnabled) {
       zoomScale(scale, y, focalPoint, options);
     }
   });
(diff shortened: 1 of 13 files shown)


src/core.js
```diff
@@ -23,50 +23,41 @@ function zoomScale(scale, zoom, center, zoomOptions) {
...
-    // based on fingers positions
-    _whichAxes = whichAxes;
-  } else {
-    // no effect
-    _whichAxes = 'xy';
-  }
-
-  const enabledScales = getEnabledScalesByPoint(zoomOptions, focalPoint.x, focalPoint.y, chart);
+
+  const xEnalbed = x !== 1 && directionEnabled(mode, 'x', chart);
```
etimberg, line 46: typo in `xEnabled`

etimberg (commented):
One minor point. The only other part I had comments about was removing some of the checks around the `enabled` option, but I'm guessing that is because they were redundant

etimberg (dismissed):

etimberg (approved):

KoalaBear84:
Thanks! Will try it out ASAP 👍

KoalaBear84:
Thanks @kurkle. Looks like it is only for 'relative' zoom/panning. I think it is not possible to do something like "zoom to last 24 hours" or so. My 'workaround' works, but everytime I call it, it shifts/moves a little. Also, it really becomes a (range) limit instead of a 'suggestion'.

See #444

KoalaBear84:
Hmm, didn't knew about that. It does work, as in that it also zooms to this period.

But it also 'locks' the graph, so you cannot zoom out, and the "Reset zoom" button I use doesn't show anymore, probably because it isn't a real zoom command anymore. Hmm, I guess I know now I see my check, it is based on the values we change on the sample you supplied.

Somehow there looks to be a change to the onZoomComplete function.

In 1.0.0-beta.1 I have to use this (with curly brackets):

```javascript
    onZoomComplete: function ({ chart }) {
        showHideZoomButton(chart);
    }

function showHideZoomButton(chart) {
	document.querySelector('.resetZoomButton').style.display = (chart.scales.x.options.min != 0 || chart.scales.x.options.max != 0 || chart.scales.y.options.min != 0 || chart.scales.y.options.max != 0) ? 'block' : 'none';
}
```

In 1.0.0-beta.2 I have to use this (without curly brackets):

```javascript
    onZoomComplete: function (chart) {
        showHideZoomButton(chart);
    }
```

Is there any logic to this, or am I doing something wrong?

PR #470, opened 2021-04-26
Combine range options, add minimum scale range

### LINKED ISSUE #396: Extreme zoom values crash the plugin
First of all many thanks for this awesome plugin ❤️ During some tests I noticed that if you zoom very far into the graph or out the plugin crashes. This is very easy to reproduce, in the first example of the documentation simply set the zoom speed to 100000 and zoom in for a few seconds.

image

image

As far as I can see the integer is overflowing and the axis disappears.

image

Would it make sense to intercept this? Or could you describe a method to define a maximum zoom level (defining for example just to allow zooming until the values are beyond three decimal places). The minmum zoom level can be set with rangeMin/rangeMax of course. Thanks a lot already!

kurkle:
Resolves #396

### DIFF
--- a/docs/guide/options.md
+++ b/docs/guide/options.md
@@ -86,4 +86,4 @@ Limits options define the limits per axis for pan and zoom.
 | ---- | -----| -----------
 | `min` | `number` | Minimun allowed value for scale.min
 | `max` | `number` | Maximum allowed value for scale.max
-| `range` | `number` | Minimum allowed range (max - min). This defines the max zoom level.
+| `minRange` | `number` | Minimum allowed range (max - min). This defines the max zoom level.
--- a/docs/guide/options.md
+++ b/docs/guide/options.md
@@ -2,7 +2,7 @@
 
 The options for chartjs-plugin-zoom should be placed in `options.plugins.zoom` in chart.js configuration.
 
-The options are split in two sub-objects, [pan](#pan) and [zoom](#zoom).
+The options are split in three sub-objects, [pan](#pan) [range](#range) and [zoom](#zoom).
 
 ```js
 const chart = new Chart('id', {
@@ -13,6 +13,9 @@ const chart = new Chart('id', {
       zoom: {
         pan: {
           // pan options and/or events
+        },
+        range: {
+          // scale range limits
         },
         zoom: {
           // zoom options and/or events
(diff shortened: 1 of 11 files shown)


etimberg (commented):
I think it makes sense and perhaps closes a few other tickets asking for zoom & pan limits. Is the options change breaking?

kurkle:
| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-zoom.esm.js` | 5.53 kB | -11 B (0%) |
| `dist/chartjs-plugin-zoom.js` | 5.78 kB | -5 B (0%) |
| `dist/chartjs-plugin-zoom.min.js` | 3.34 kB | -8 B (0%) |

kurkle:
> I think it makes sense and perhaps closes a few other tickets asking for zoom & pan limits. Is the options change breaking?

It is, added the label.

etimberg (approved):

docs/guide/options.md
```diff
@@ -69,3 +68,22 @@ const chart = new Chart('id', {
...
+## Range
+
+Range options define the limits for scale pan and zoom.
+
+### Range options
+
+| Name | Type | Description
+| ---- | -----| -----------
+| `x` | [`ScaleRange`](#scale-range) | Range limits for x-axis
+| `y` | [`ScaleRange`](#scale-range) | Range limits for x-axis
```
jledentu, line 81: typo: y-axis

docs/guide/options.md
```diff
@@ -69,3 +68,22 @@ const chart = new Chart('id', {
...
+
+#### Scale Range
+
+| Name | Type | Description
+| ---- | -----| -----------
+| `min` | `number` | Minimun allowed value for scale.min
+| `max` | `number` | Maximum allowed value for scale.max
+| `range` | `number` | Minimum allowed range (max - min). This defines the max zoom level.
```
jledentu, line 89: Maybe it's just me but `range.range` naming is a bit confusing, and doesn't reflect that defines a limit. ^^ What do you think about `minRange` or `limit`... ?

jledentu (commented):
Great! This `range` limit will be very useful! :+1:

docs/guide/options.md
kurkle: Changed the option key to `limits`, so `limits.min`, `limits.max` and `limits.range`. What do you think?

kurkle: Maybe it should still be `minRange` to leave room for possible `maxRange`.

jledentu: As you want, I think that `limits.min`, `limits.max` and `limits.range` is nice! Thanks!

jledentu (approved):

etimberg (approved):

kurkle:
Last change to `minRange`, so we can have `maxRange` in the future if needed.

etimberg (approved):

Issue #169, opened 2018-10-04
Allow extending the zoom plugin for custom scales


wheercool:
Hi, I'm looking for ability to inject custom scale, but it looks like it is impossible.

I can do it only manually adding my scale to the list of zoom/panFunctions.

```
zoomNS.zoomFunctions.category = zoomIndexScale;
zoomNS.zoomFunctions.time = zoomTimeScale;
zoomNS.zoomFunctions.linear = zoomNumericalScale;
zoomNS.zoomFunctions.duration = zoomNumericalScale;  // this was added
zoomNS.zoomFunctions.logarithmic = zoomNumericalScale;
zoomNS.panFunctions.category = panIndexScale;
zoomNS.panFunctions.time = panTimeScale;
zoomNS.panFunctions.linear = panNumericalScale;
zoomNS.panFunctions.duration = panNumericalScale; // this was added
zoomNS.panFunctions.logarithmic = panNumericalScale;
```
My custom duration scale works almost exactly like a linear. Is there any solution I can inject my scale w/o touching source code?

jledentu:
Yeah, I don't think you can do that for the moment. In the future, we may consider a scale as numerical by default?

jstsddg:
I found a hack which allowed me to add custom scales without touching the source code. But as this is not really a clean solution, nor a documented one, I think adding some options would be benefitial.
```JavaScript
Chart.Zoom.panFunctions.myCustomNumericalScale = Chart.Zoom.panFunctions.linear;
Chart.Zoom.zoomFunctions.myCustomNumericalScale = Chart.Zoom.zoomFunctions.linear;

Chart.Zoom.panFunctions.myCustomCategoryScale = Chart.Zoom.panFunctions.category;
Chart.Zoom.zoomFunctions.myCustomCategoryScale = Chart.Zoom.zoomFunctions.category;
```

kurkle:
From v1.0.0-beta.3 numeric functions are used as default, if no scale type specific function is found. So these cases would be handled by that.

kurkle:
The objects are now available as of #481, but the documentation is missing.

PR #481, opened 2021-04-30
Expose zoomFunctions and panFunctions for customization

nagix:
I'm tying to upgrade [chartjs-plugin-streaming](https://github.com/nagix/chartjs-plugin-streaming) for Chart.js 3.x and chartjs-plugin-zoom 1.x, but the latest code doesn't expose `zoomFunctions` or `panFunctions` which are required to support new scale types. So, this PR adds these internal objects to the plugin object as properties.


nagix:
#169 is related to this

kurkle (approved):

PR #483, opened 2021-04-30
Make resetZoom accept UpdateMode argument

nagix:
This PR makes `chart.resetZoom` accept the [`UpdateMode`](https://www.chartjs.org/docs/master/api/enums/updatemodeenum.html) argument so that application developers can control animations when they call it.


kurkle (approved):
Missed that one, thanks

Issue #373, opened 2020-07-08
Add callback for the start of pan/zoom


Remagpie:
I'm currently working on a component that changes the internal state when the user is panning or zooming.
I implemented it using `onPan` and `onZoom` for now, but it'd be nice if there is an callback that's called only at the start of the panning or zooming just like the `onPanComplete` and `onZoomComplete`.

ZaLiTHkA:
I've been working on a chart myself where `onPanStart` and `onZoomStart` would have been useful.. unfortunately, I'm currently chasing deadlines for this project, so I can't do anything now, but (unless someone else beats me to it) I might just fork this plugin and look at implementing these myself this coming weekend. (:

Issue #276, opened 2019-10-08
Pan start region


albmac:
Is it possible to restrict the starting point of a pan action to a certain region of the chart (like only around the middle)?
My problem is that in order to be readable on a smartphone, a chart should have full width, but then scrolling above it triggers a pan.
Having instead a smaller "pan start region" would allow both scrolling and panning.
Is this something you can add in the future?

albmac:
Meanwhile I found a workaround: place some transparent rectangles above the chart where it should not be responsive to pan/zoom:
<div class="chart-container">
<canvas id="mycanvas"></canvas>
<div style="position:absolute;left:0;top:0;width:25%;height:100%;background:rgba(0,0,0,0);"></div>
<div style="position:absolute;right:0;top:0;width:25%;height:100%;background:rgba(0,0,0,0);"></div>
</div><br>
Now it will only see the mouse/finger in the central region (25% to 75%)

JZubero:
@albmac Kinda workaroundy but does the trick. This was working for me as well.

PR #487, opened 2021-05-01
Add onPanStart and onZoomStart callbacks

kurkle:
Closes: #373
Closes: #276

* add onPanStart / onZoomStart handlers, that can return `false` to cancel the action
* add sample for limiting pan start to middle of chart
* add tests
* fix bug when `delta` is undefined
* fix callback typings

### DIFF
--- a/docs/guide/options.md
+++ b/docs/guide/options.md
@@ -45,6 +45,7 @@ const chart = new Chart('id', {
 | `onPan` | `{chart}` | Called while the chart is being panned
 | `onPanComplete` | `{chart}` | Called once panning is completed
 | `onPanRejected` | `{chart,event}` | Called when panning is rejected due to missing modifier key. `event` is the a [hammer event](https://hammerjs.github.io/api#event-object) that failed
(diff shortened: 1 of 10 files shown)


docs/guide/options.md
```diff
@@ -45,6 +45,7 @@ const chart = new Chart('id', {
 | `onPan` | `{chart}` | Called while the chart is being panned
 | `onPanComplete` | `{chart}` | Called once panning is completed
 | `onPanRejected` | `{chart,event}` | Called when panning is rejected due to missing modifier key. `event` is the a [hammer event](https://hammerjs.github.io/api#event-object) that failed
+| `onPanStart` | `{chart,event,point}` | Called when panning is about to start. If this callback returns true, panning is aborted and `onPanRejected` is invoked
```
etimberg, line 48: In the core we use return false to abort, should we be consistent here?

kurkle: Good point

etimberg (approved):

PR #500, opened 2021-05-06
Cleans the click to zoom sample MD file

stockiNail:
Cleans the click to zoom sample MD file because there was a sample text


kurkle:
The text is there on purpose, so the page is scrollable. This demoes that the chart does not consume the wheel until clicked.
(The chart editor consumes it after a little delay though)

Maybe it would be good to add a note about that :)

stockiNail:
I am very sorry. I didn't recognise that purpose. It was weird.
I close this PR and I'm going to submit another one with a note below the title of the page.

Issue #239, opened 2019-05-21
When drag zoom is enabled, scroll zooming is blocked


gornad:
I am new to this plugin. When adding drag: true, that disables scroll zooming. Is it not possible to have scroll zooming and drag zooming at the same time? This can be tested by simply adding drag: true to this http://jsfiddle.net/jyougo/sv2n4zdh/

benmccann:
I would be open to a PR to allow both at the same time

mshariq-nerd:
@gornad @benmccann  Any solution for this? I am also looking for the same functionality. Any workaround?

gornad:
Line 414 (ADDED  && event.which === 3):
 if (chartInstance.$zoom._options.zoom && chartInstance.$zoom._options.zoom.drag && event.which === 3) {

Line 469 (CHANGED FROM (dragDistanceX > 0 || dragDistanceY > 0)):
if (dragDistanceX > 5 && dragDistanceY > 0) {

Line 480 (CHANGED FROM !(chartInstance.$zoom._options.zoom && chartInstance.$zoom._options.zoom.drag)): 
if (chartInstance.$zoom._options.zoom) {

These are the changes I made in the library. I didn't make a pull request because I don't know how to properly contribute to projects on github. Feel free to take these changes and merge them yourself.

benmccann:
Right now there's just a single option `drag` which enables drag and disables scroll. I agree it would make sense to be able to control these separately. We would also need a new user option added to control the `scroll` in that case. I would be open to a PR to add such a feature

jledentu:
@benmccann What would you think about a such configuration?

```js
zoom: {
    // Boolean to enable zooming
    enabled: true,

    interaction: ['scroll', 'drag'], // Enable zoom by scroll, or drag, or both

    // Drag-to-zoom effect can be customized
    dragOptions: {
        borderColor: 'rgba(225,225,225,0.3)'
        borderWidth: 5,
        backgroundColor: 'rgb(225,225,225)',
        animationDuration: 0
    },

    // Scroll can be customized
    scrollOptions: {
        // Speed of zoom via mouse wheel
        // (percentage of zoom on a wheel event)
        speed: 0.1,
    }

    // Zooming directions. Remove the appropriate direction to disable
    // Eg. 'y' would only allow zooming in the y direction
    // A function that is called as the user is zooming and returns the
    // available directions can also be used:
    //   mode: function({ chart }) {
    //     return 'xy';
    //   },
[...]

PR #507, opened 2021-05-12
Add the ability to enable both wheel and drag modes at the same time

jledentu:
Hi,

Here is a PR thats adds the ability to enable both wheel and drag modes at the same time, as suggested in #239. I propose to split the wheel/drag configurations like this:

```js
zoom: {
      wheel: {
        enabled: true,
        speed: 0.1,
        modifierKey: null
      },
      drag: {
        enabled: true,
        backgroundColor: 'yellow',
        borderColor: 'black',
        borderWidth: 1,
        threshold: 1
      },
      mode: 'xy',
}
```

`speed`, `threshold`, `wheelModifierKey` (now `wheel.modifierKey`) are moved to their respective configurations, I think that makes clearer that a given option refers to a given mode.

### DIFF
--- a/docs/guide/options.md
+++ b/docs/guide/options.md
@@ -53,21 +53,28 @@ const chart = new Chart('id', {
 
 | Name | Type | Default | Description
 | ---- | ---- | ------- | ----------
-| `enabled` | `boolean` | `false` | Enable zooming
-| `drag` | `boolean`\|[`DragEffectOptions`](#drag-effect-options) | `undefined` | Enable drag-to-zoom behavior (disables zooming by wheel)
+| `wheel` | [`WheelOptions`](#wheel-options) | `undefined` | Options of the mouse wheel behavior
+| `drag` | [`DragOptions`](#drag-options) | `undefined` | Options of the drag-to-zoom behavior
 | `mode` | `'x'`\|`'y'`\|`'xy'` | `'xy'` | Allowed zoom directions
 | `overScaleMode` | `'x'`\|`'y'`\|`'xy'` | `undefined` | Which of the enabled zooming directions should only be available when the mouse cursor is over a scale for that axis
-| `speed` | `number` | `0.1` | Factor of zoom speed via mouse wheel.
-| `threshold` | `number` | `0` | Mimimal zoom distance required before actually applying zoom
-| `wheelModifierKey` | `'ctrl'`\|`'alt'`\|`'shift'`\|`'meta'` | `null` |  Modifier key required for zooming with mouse
 
-#### Drag effect options
+#### Wheel options
 
-| Name | Type | Description
-| ---- | -----| -----------
-| `backgroundColor` | `Color` | Fill color
-| `borderColor` | `Color` | Stroke color
-| `borderWidth` | `number` | Stroke width
+| Name | Type | Default | Description
+| ---- | -----| ------- | -----------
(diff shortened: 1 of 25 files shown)


kurkle:
Nice!
How about `pinch` zoom? I think its still checking for `zoom.enabled`. Maybe it should be possible to disable that separately? Own scope for it?

jledentu:
Oh you're right, I admit that I often forget to check touch behaviors. Do I add `pinch.enabled`? I guess that it will often be aligned with `wheel.enabled`  as I don't know if it makes sense to explicitly enable zoom by mouse and disable zoom on touch devices. But at least we allow more flexibility.

kurkle:
I think `pinch.enabled` would be good. Maybe fall back from each scope to `zoom.enabled`, to make the "break" of the change smaller and maybe easier to find.

kurkle (dismissed):

jledentu:
@kurkle I added `pinch.enabled` + a warning on `zoom.enabled ` as suggested (I think it's better to remove this option in order to keep the code clean and make the configuration more explicit?).

docs/guide/options.md
```diff
@@ -53,21 +53,35 @@ const chart = new Chart('id', {
 
 | Name | Type | Default | Description
 | ---- | ---- | ------- | ----------
-| `enabled` | `boolean` | `false` | Enable zooming
-| `drag` | `boolean`\|[`DragEffectOptions`](#drag-effect-options) | `undefined` | Enable drag-to-zoom behavior (disables zooming by wheel)
+| `wheel` | [`WheelOptions`](#wheel-options) | `undefined` | Options of the mouse wheel behavior
```
etimberg, line 56: Should make sure this is in the release notes as a breaking change. Or we can keep `enabled` at the top level and do some kind of fallback

src/plugin.js
```diff
@@ -19,17 +19,29 @@ export default {
...
-      wheelModifierKey: null
     }
   },
 
   start: function(chart, _args, options) {
     const state = getState(chart);
     state.options = options;
 
+    if (Object.prototype.hasOwnProperty.call(options.zoom, 'enabled')) {
+      console.warn('The option `zoom.enabled` is not more supported. Please use `zoom.wheel.enabled`, `zoom.drag.enabled`, or `zoom.pinch.enabled`.');
```
etimberg, line 42: I would say "no longer supported."

etimberg (commented):
Two very minor comments. Looks great!

src/plugin.js
jledentu: You're right, fixed. :+1:

etimberg (approved):

docs/guide/options.md
jledentu: I think it's the good moment to make breaking changes since the final 1.0.0 has not been released yet (I guess there were other breaking changes in previous beta releases but I'm not sure)? I'm afraid that supporting a fallback could bring some confusion when a specific behavior works without having been explicitly enabled?

+ the warning in the console will make migration easier for users who didn't read the release notes. 

But yeah, I agree that a mention in the release note is really important. I can edit the release draft if you want.

kurkle: @jledentu just in case you hadn't noticed, release drafter rewrites the notes after every commit to master, so should do the edit after the last commit (bump) before release.

kurkle (approved):

Issue #523, opened 2021-05-25
Option to disable zooming out


joshkel:
I have what I thought would be a common use case, so I'm not sure if I'm missing something: My charts are initially displayed showing all available data, so there's no reason to zoom out. And allowing the user to arbitrarily zoom out doesn't feel like a very good UX; it results in a tiny plot with no good way to exactly zoom back in to the data of interest.

I realize that I can use `limits` properties to control this behavior, but that doesn't work if I'm relying on Chart.js to automatically select scale limits or to select adjusted scale limits based on `suggestedMin` and `suggestedMax`.

As a feature request, could chartjs-plugin-zoom add an option to disable zooming out? Maybe a ScaleLimits `disableZoomOut` flag or some sort of `useOriginalMin` / `autoMin` and ` useOriginalMax` / `autoMax` flags?

PR #524, opened 2021-05-28
Multi-axis zoom and original zoom limits

### LINKED ISSUE #522: Zoom limits with multiple axes
I'm using chartjs-plugin-zoom on a chart with multiple axes and am having trouble configuring limits. The chart correctly handles zooming multiple axes, but because the `limits` object can only be defined for a single x and a single y, there's no way to configure different zoom limits for different axes.

See https://codesandbox.io/s/zen-ishizaka-ztujp?file=/src/index.js for an example.

Suggestion: Change the `limits` object so that, in addition to the default `x` and `y`, it allows ScaleLimits objects indexed by the scale ID. If this seems reasonable, I can submit a PR.

joshkel:
This fixes #522 and #523. I put it together for my own use, so everything's in one branch, but I'm happy to submit as smaller PRs for discussion and review whenever someone's ready to take a look at it.

Thank you.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -34,7 +34,7 @@ function getCenter(chart) {
 /**
  * @param chart The chart instance
  * @param {number | {x?: number, y?: number, focalPoint?: {x: number, y: number}}} amount The zoom percentage or percentages and focal point
- * @param {string} [transition] Which transiton mode to use. Defaults to 'none'
+ * @param {string} [transition] Which transition mode to use. Defaults to 'none'
  */
 export function zoom(chart, amount, transition = 'none') {
   const {x = 1, y = 1, focalPoint = getCenter(chart)} = typeof amount === 'number' ? {x: amount, y: amount} : amount;
--- a/src/core.js
+++ b/src/core.js
@@ -8,7 +8,10 @@ function storeOriginalScaleLimits(chart) {
   const {originalScaleLimits} = chartState;
   each(chart.scales, function(scale) {
     if (!originalScaleLimits[scale.id]) {
-      originalScaleLimits[scale.id] = {min: scale.options.min, max: scale.options.max};
+      originalScaleLimits[scale.id] = {
+        min: {scale: scale.min, options: scale.options.min},
+        max: {scale: scale.max, options: scale.options.max},
+      };
     }
   });
   each(originalScaleLimits, function(opt, key) {
--- a/src/core.js
+++ b/src/core.js
@@ -3,8 +3,8 @@ import {panFunctions, updateRange, zoomFunctions} from './scale.types';
 import {getState} from './state';
 import {directionEnabled, getEnabledScalesByPoint} from './utils';
 
-function storeOriginalScaleLimits(chart) {
-  const {originalScaleLimits} = getState(chart);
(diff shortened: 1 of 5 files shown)


kurkle:
Had a quick look, looks good. The .gitignore changes should be split off, if those are needed at all. Not a fan of passing the state around that much (actually just the number of args), but I think its better than getting the state everywhere.

kurkle:
Something is broken:

>[test-karma]       ✗ /base/test/fixtures/zoom/zoom-reset.js
[test-karma] 	RangeError: invalid digits value: NaN thrown

joshkel:
Thanks. I fixed the test failure.

I wasn't a big fan of the number of parameters, either, but I didn't see a clean way of reducing it.

I noticed that the module exports zoomFunctions and panFunctions, so that third-party custom scale classes can register specialized handling for zoom and pan (#169). This PR would change the function signature of those functions, which would be a SemVer breaking change. Should I move the chartState parameter to the end of each of the zoomFunctions and panFunctions to avoid this? Or, since the issue is still open and the feature is not yet documented, is it okay to make changes?

kurkle:
Good catch the breaking change on `zoomFunctions` and `panFunctions`.

Maybe the extra parameters should be avoided at least on those functions, as the `getState` should be fast enough to just call again in there.

src/core.js
```diff
@@ -34,24 +38,24 @@ function getCenter(chart) {
 /**
  * @param chart The chart instance
  * @param {number | {x?: number, y?: number, focalPoint?: {x: number, y: number}}} amount The zoom percentage or percentages and focal point
- * @param {string} [transition] Which transiton mode to use. Defaults to 'none'
+ * @param {string} [transition] Which transition mode to use. Defaults to 'none'
  */
 export function zoom(chart, amount, transition = 'none') {
   const {x = 1, y = 1, focalPoint = getCenter(chart)} = typeof amount === 'number' ? {x: amount, y: amount} : amount;
   const {options: {limits, zoom: zoomOptions}} = getState(chart);
```
kurkle, line 45: This getState call can be removed and the things read from the `chartState` returned from `storeOriginalScaleLimits`

src/core.js
```diff
@@ -73,15 +77,15 @@ export function zoomRect(chart, p0, p1, transition = 'none') {
...
```
kurkle, line 77: Same here. Might actually be better to pass the state to storeOriginalScaleLimits?

joshkel:
Thanks for the feedback.

If I avoid the extra parameter for `zoomFunctions` and `panFunctions` and instead retrieve the state within the `updateRange` function that they call, then most of the changes that I'd originally made end up being unnecessary.

So it's probably still a bit cleaner to call `getState` once within the top-level public API methods, and use it in places like [this](https://github.com/chartjs/chartjs-plugin-zoom/pull/524#discussion_r643164153), and pass it to `storeOriginalScaleLimits`, but it's no longer needed for this PR. Would you like me to go ahead and do it?

kurkle:
I'd be ok having it done in this pr, if you are willing to do it ;)

joshkel:
@kurkle Done. Thanks.

kurkle (approved):

PR #536, opened 2021-06-06
Add getZoomLevel function

kurkle:
Changes:
 * new `getZoomLevel` function
 * `onZoomComplete` call after `resetZoom`
 * rename `chartState` to `state` for consistency
 * utilize `valueOrDefault` - this could break something, it is not checking for `null`. @joshkel was there a reason for checking `null`?

### DIFF
--- a/docs/samples/basic.md
+++ b/docs/samples/basic.md
@@ -68,12 +68,18 @@ const zoomOptions = {
       enabled: true
     },
     mode: 'xy',
+    onZoomComplete({chart}) {
+      // This update is needed to display up to date zoom level in the title.
+      // Without this, previous zoom level is displayed.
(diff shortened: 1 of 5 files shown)


joshkel:
I don't think checking for null is needed; I was mostly being paranoid / overly defensive.

If I'm following the Chart.js code correctly, null can show up for a scale's min and max (see [here](https://github.com/chartjs/Chart.js/blob/6a46cc39bbf51f4ae2f7d716efe9ec450390a0bf/src/scales/scale.logarithmic.js#L74)), however, even if it did, Math.min and Max.max will ignore it, which seems like the desired behavior.

etimberg (approved):

Issue #533, opened 2021-06-01
Chart instances persist in memory after destroy


Viceriel:
Hello, I recently switched Chart.js from v2.x to 3.x. I also use the chartjs-plugin-zoom 1.0.1. My situation is this I have SPA with some tabs, on one tab are the charts, between them line chart and for this one I use the zoom plugin, because data are usually huge.

When I transit from page with charts to another, this is called for cleaning:

`if (this._chart)`
`{`
    `this._chart.options.onClick = null;`
   `//desperate attempt to save situation`
    `this._chart.options.plugins = null;`
    `this._chart.clear();`
    `this._chart.destroy();`
`}`

I expected that destroy do everything necessary from memory perspective. For version 2 it is working, but here I observe strange behavior. When I switch from chart page to another and inspect memory I saw LineController(object asociated with chart.js) between objects. When I switch back to chart page and then again to another memory consumtion raise and count of LineController too. Raising instances

What i find out is this, when I comment the zoom plugin from options everything is working again, no more raising LineControll instances no visible memory consumption raise between pages switching. But this plugin is necessary for me, so commenting out is not good option. I tried manually set plugins for chart to null:

`this._chart.options.plugins = null;`
But this not solve the memory problem. Has anybody face such condition? I also posted question to with code snippets [so](https://stackoverflow.com/questions/67794561/chart-js-v3-after-using-of-zoom-plugin-instance-persist-in-memory)

kurkle:
Are you possibly using any polyfills?

Viceriel:
At least I'm not aware of any polyfills. When I find moment to take a breath I try to extract minimal example to flems(but not sure, how such platform allows to inspect memory, so maybe minimal html like samples is more appropriate?).

Viceriel:
Hello. I tried extract minimal example. So I not observe same behavior(no more LineController). On the other hand seems like everything else is same. Memory consumption raises despite destroy called and only if I use plugin zoom. It looks like this, first sample is desperate toggling between pages after loading the site. Second sample is desperate toggling between pages after first one.

![desperaterun](https://user-images.githubusercontent.com/5782564/120446435-d7990e00-c389-11eb-8276-28023b372e8e.jpg)

PR #541, opened 2021-06-07
Fix removing event listeners

joshkel:
The plugin only removed its event listeners if it could access the chart's canvas.  However, Chart.js clears its canvas before notifying plugins, so this didn't work:

https://github.com/chartjs/Chart.js/blob/v3.3.2/src/core/core.controller.js#L857

I noticed this when I destroyed and recreated a Chart.js object against the same HTML `<canvas>` element; the original chart's event handlers still fired and threw errors because the `getState` object had been removed and no longer had needed information.

To address this, I'm now saving the event listener's target as a custom property on the event listener function; that way, `removeHandler` always knows exactly what the target is.

I also noticed that `mouseUp` was calling `removeHandler` with incorrect parameters, so I fixed it.

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -3,19 +3,20 @@ import {zoom, zoomRect} from './core';
 import {callback as call} from 'chart.js/helpers';
 import {getState} from './state';
 
-function removeHandler(chart, target, type) {
+function removeHandler(chart, type) {
   const {handlers} = getState(chart);
   const handler = handlers[type];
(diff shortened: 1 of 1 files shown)


kurkle (approved):

kurkle:
Fixes #533?

joshkel:
From a simple local test (with only basic plot features enabled), yes, it looks like it fixes #533.

PR #556, opened 2021-06-30
Add a modifierKey for drag-to-zoom

Sean-Stellingwerff:
This PR adds the `modifierKey` option to `zoom.drag`, in a similar fashion to how it works for panning and wheel zooming.

Note: by some hilarious coincidence it looks like @kurkle had the exact same idea today and beat me by 4 hours to add this 😂 I noticed that I had a couple of changes not included in his PR, so I've decided to add my MR as well. I'm perfectly fine with either PR getting merged: I care more about the feature being there than getting my code merged 😉.

### DIFF
--- a/src/hammer.js
+++ b/src/hammer.js
@@ -4,18 +4,32 @@ import {pan, zoom} from './core';
 import {getState} from './state';
 import {directionEnabled, getEnabledScalesByPoint} from './utils';
 
+function blockedByPanModifierKey(panOptions, event) {
+  const modifierKey = panOptions.modifierKey;
+  const requireModifier = modifierKey && (event.pointerType === 'mouse');
+  return requireModifier && !event.srcEvent[modifierKey + 'Key'];
+}
+
+function blockedByDragModifierKey(zoomOptions, event) {
+  if (!zoomOptions || !zoomOptions.drag || !zoomOptions.drag.enabled) {
+    return false;
+  }
+  const modifierKey = zoomOptions.drag.modifierKey;
+  const requireModifier = modifierKey && (event.pointerType === 'mouse');
+  return requireModifier && event.srcEvent[modifierKey + 'Key'];
+}
+
 function createEnabler(chart, state) {
   return function(recognizer, event) {
     const panOptions = state.options.pan;
+    const zoomOptions = state.options.zoom;
     if (!panOptions || !panOptions.enabled) {
       return false;
     }
     if (!event || !event.srcEvent) { // Sometimes Hammer queries this with a null event.
       return true;
     }
-    const modifierKey = panOptions.modifierKey;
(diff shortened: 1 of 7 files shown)


kurkle:
| Filename | Size | Change |
| :--- | :---: | :---: |
| `dist/chartjs-plugin-zoom.esm.js` | 7.46 kB | +106 B (+1%) |
| `dist/chartjs-plugin-zoom.js` | 7.71 kB | +108 B (+1%) |
| `dist/chartjs-plugin-zoom.min.js` | 4.49 kB | +67 B (+2%) |

test/specs/zoom.spec.js
```diff
@@ -396,6 +396,79 @@ describe('zoom', function() {
     }
   });
 
+  describe('drag with zoom.drag.modifierKey', function() {
+    for (const key of ['ctrl', 'alt', 'shift', 'meta']) {
+      for (const pressed of [true, false]) {
+        let chart, scaleX, scaleY;
+        it(`should ${pressed ? 'not ' : ''}change ${pressed ? 'without' : 'with'} key ${key}`, async function() {
```
kurkle, line 403: The 'not'/'' and 'without'/'with' are reversed

src/hammer.js
```diff
@@ -4,18 +4,32 @@ import {pan, zoom} from './core';
...
     if (!panOptions || !panOptions.enabled) {
       return false;
     }
     if (!event || !event.srcEvent) { // Sometimes Hammer queries this with a null event.
       return true;
     }
-    const modifierKey = panOptions.modifierKey;
-    const requireModifier = modifierKey && (event.pointerType === 'mouse');
-    if (!state.panning && requireModifier && !event.srcEvent[modifierKey + 'Key']) {
+    if (!state.panning && blockedByPanModifierKey(panOptions, event) || blockedByDragModifierKey(zoomOptions, event)) {
```
kurkle, line 32: I think the `!state.panning` check is needed for both conditions; once panning is started, it should not change to drag-zoom if that modifier key is pressed while still panning.

kurkle (commented):
I left a couple of comments.

The other pr is about half the size, but we are still talking about couple of bytes. I ofc prefer the way I did it :)
However, its easier to get this merged as I don't need to request a review from another maintainer.

I also noticed that `pan.enabled` can be false and a modifierKey in there would still prevent drag-to-zoom. I changed that in the other pr.

test/specs/zoom.spec.js
Sean-Stellingwerff: Whoops, good catch, thanks!

Sean-Stellingwerff:
Thanks for the feedback @kurkle!

I would just go with your PR. It's the cleaner of the two, it's smaller and at least for me there's no immediate rush to get it merged.

PR #585, opened 2021-09-17
Add chart method for getting initial scale bounds

MrJohz:
Based on the discussion in #584, this adds a new method, `getInitialScaleBounds` that returns an object mapping scale IDs to {min, max} values.

```js
console.log(chart.getInitialScaleBounds());
// {
//    x: {min: 3, max: 8},
//    y: {min: -10, max: 10}
// }
```

I've added tests and type definitions.  I couldn't see anywhere obvious to document this in the guide, but I think exporting it from `index.d.ts` at least adds some basic type doc stuff, right?

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -212,3 +212,20 @@ export function getInitialScaleBounds(chart) {
 
   return scaleBounds;
 }
+
+export function isZoomedOrPanned(chart) {
+  const scaleBounds = getInitialScaleBounds(chart);
+  for (const scaleId of Object.keys(chart.scales)) {
+    const {min: originalMin, max: originalMax} = scaleBounds[scaleId];
+
+    if (chart.scales[scaleId].min !== originalMin) {
+      return true;
+    }
+
+    if (chart.scales[scaleId].max !== originalMax) {
+      return true;
+    }
+  }
+
+  return false;
+}
--- a/src/core.js
+++ b/src/core.js
@@ -202,3 +202,18 @@ export function pan(chart, delta, enabledScales, transition = 'none') {
   call(onPan, [{chart}]);
 }
 
+export function getInitialScaleBounds(chart) {
+  const state = getState(chart);
+  const scaleBounds = {};
+  for (const scaleId of Object.keys(chart.scales)) {
+    const original = state.originalScaleLimits[scaleId];
+    if (!original) {
+      continue;
+    }
+
+    const {min, max} = original;
+    scaleBounds[scaleId] = {min: min.scale, max: max.scale};
+  }
+
+  return scaleBounds;
+}
--- a/src/core.js
+++ b/src/core.js
@@ -206,12 +206,7 @@ export function getInitialScaleBounds(chart) {
   const state = getState(chart);
   const scaleBounds = {};
   for (const scaleId of Object.keys(chart.scales)) {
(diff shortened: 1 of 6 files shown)


LeeLenaleee:
Guess to keep it in line with the main chartjs docs it's best to create a sub cat under the developer section call it API and put it in there

https://www.chartjs.org/docs/master/developers/api.html

src/core.js
```diff
@@ -202,3 +202,18 @@ export function pan(chart, delta, enabledScales, transition = 'none') {
...
+  const scaleBounds = {};
+  for (const scaleId of Object.keys(chart.scales)) {
+    const original = state.originalScaleLimits[scaleId];
+    if (!original) {
+      continue;
+    }
+
+    const {min, max} = original;
+    scaleBounds[scaleId] = {min: min.scale, max: max.scale};
+  }
```
kurkle, line 216: Maybe the `undefined` values would be valuable too, when nothing is found?

```suggestion
  for (const scaleId of Object.keys(chart.scales)) {
    const {min, max} = state.originalScaleLimits[scaleId] || {min: {}, max: {}};
    scaleBounds[scaleId] = {min: min.scale, max: max.scale};
  }
```

kurkle:
Added a suggestion, but I did not try this - so it might not make sense.
Looks good to me. Wondering if a (additional?) higher level `isZoomedOrPanned()` (or something) would be a better API for this kind of use cases?

src/core.js
MrJohz: I was ignoring that case for now, because I assumed that it shouldn't be possible — if the scale exists on the chart, it should also exist on `originalScaleLimits`, I _think_.  But I'm not certain I understand all the states that this can be in, so maybe there's a weird edge case where this becomes important!

MrJohz:
> Guess to keep it in line with the main chartjs docs it's best to create a sub cat under the developer section call it API and put it in there

:+1: Can do

> Looks good to me. Wondering if a (additional?) higher level `isZoomedOrPanned()` (or something) would be a better API for this kind of use cases?

I can imagine getting the original bounds will still be useful, but you're right, having the higher level API there as well would also be useful.  I'll have a go at adding that as well, although I might only get round to that on Monday or something.

life-of-e:
A isZoomedOrPanned() might be helpful, it may also address my problem where trying to reset all of the charts on a page at once breaks (see my Discussion #583)

MrJohz:
I've added documentation for all methods added on the Chart.js instance.  I also noticed while doing that, that the `zoom` method had the wrong Typescript types, so I fixed that as well.

@kurkle, does that work for you?

kurkle (approved):
Nice work, thank you!

I've requested a review from @etimberg too, lets see if he has any notes

kurkle:
Looks like there are some linting errors

MrJohz:
I think I've fixed that issue, there was also a linting error in the documentation somewhere that I believe I've fixed now as well.

kurkle (approved):
coveralls.io is in maintenance currently, that's why the CI fails. I'd expect it to become available later today.

etimberg (approved):

MrJohz:
@kurkle, looks like it's back up and running again.  Is there an easy way to rerun the failed jobs, or do I need to push a new commit or something?

PR #596, opened 2021-11-16
Documentation versioning

kurkle:


etimberg (dismissed):

kurkle:
Looks like the executable bit can not be set while adding a file.

etimberg (approved):

PR #630, opened 2022-03-10
Fix initial call to isZoomedOrPanned

### LINKED ISSUE #629: isZoomedOrPanned incorrectly returns true on initial render
isZoomedOrPanned incorrectly returns true when a chart is first rendered, before the chartjs-plugin-zoom plugin does anything.

Steps to reproduce: See https://codesandbox.io/s/is-zoomed-xo90ep.

1. Click the "Check zoom" button and note that it incorrectly reports true.
2. Click and drag somewhere on the chart. This invokes the chartjs-plugin-zoom plugin by trying to do a pan operation, but nothing happens, because the chart is already at its limits.
3. Click the "Check zoom" button and note that it correctly reports false.

joshkel:
isZoomedOrPanned incorrectly returns true if it's called before any zoom or pan operations are done.

I believe that the `!== undefined` check should be valid; storeOriginalScaleLimits always sets the min/max value to chart.scales[scaleId]'s min/max, which TypeScript says are always defined. However, I could be missing something.

Fixes #629

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -218,11 +218,11 @@ export function isZoomedOrPanned(chart) {
   for (const scaleId of Object.keys(chart.scales)) {
     const {min: originalMin, max: originalMax} = scaleBounds[scaleId];
 
-    if (chart.scales[scaleId].min !== originalMin) {
+    if (originalMin !== undefined && chart.scales[scaleId].min !== originalMin) {
       return true;
     }
 
-    if (chart.scales[scaleId].max !== originalMax) {
+    if (originalMax !== undefined && chart.scales[scaleId].max !== originalMax) {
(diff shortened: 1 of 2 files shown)


kurkle (approved):

joshkel:
I see that the build failed due to excessive cognitive complexity. Do you want me to fix this? I noticed that CodeClimate's cognitive complexity limit of 5 seems stricter than the configured ESLint cyclomatic complexity limit of 10. I don't see obvious unnecessary complexity, but I'm happy to break apart the function to reduce the complexity score if that's what's desired.

kurkle:
No, I don't think that needs to be changed (approved it)

PR #634, opened 2022-03-23
Publish latest/next docs in separate job

### LINKED ISSUE #606: latest/next docs tags will break in next release
https://github.com/chartjs/chartjs-plugin-annotation/issues/565

kurkle:
Fix #606


etimberg (approved):
I think it looks good, but unsure how to test it

kurkle:
> I think it looks good, but unsure how to test it

Probably only testable by releasing. I tried to make it identical with annotation plugin and there it works.

PR #658, opened 2022-05-20
Replace overScaleMode with more flexible scaleMode

joshkel:
As described at https://github.com/chartjs/chartjs-plugin-zoom/discussions/612, I'd like to allow both regular zooming (to zoom both axes via mouse wheel or pinch on the main chart area) and zooming a specific scale (to zoom only one axis via mouse wheel or pinch over the scale axis), but it seems that `overScaleMode` prevents regular zooming from working.

I would have expected that `mode` control regular zooming, while `overScaleMode` controls scale zooming in particular. For example ("scenario A"):

* `{mode: 'xy', overScaleMode: 'y'}` would mean that zooming over the regular chart scales X and Y, and zooming over the Y axis scales Y.
* `{mode: 'x', overScaleMode: 'y'}` would mean that zooming over the regular chart scales X, and zooming over the Y axis scales Y.
* `{mode: 'xy', overScaleMode: 'xy'}` would mean that zooming over the regular chart scales X and Y, and zooming over the X or Y axis scales X or Y, respectively.

Instead, it appears that `overScaleMode` restricts and modifies `mode` ("scenario B"):

* `{mode: 'xy', overScaleMode: 'y'}` would mean that zooming over the regular chart scales only X, and zooming over the Y axis scales Y.
* `{mode: 'x', overScaleMode: 'y'}` is an invalid configuration.
* `{mode: 'xy', overScaleMode: 'xy'}` would mean that zooming over the regular chart does nothing, and zooming over the X or Y axis scales X or Y, respectively.

This PR adds a new property, `scaleMode`, that behaves like scenario A. Since scenario A is strictly more flexible than scenario B, it deprecates `overScaleMode`.

There is a potential change in behavior with this PR: Previously, the `pan` function used `chart.scales` if `overScaleMode` was undefined, and it evaluated `directionEnabled` (which called `mode` if `mode` was a function). This meant that it could handle changes to the `mode` function's return value, and changes to what chart scales exist, even in the middle of a pan operation. Now, the list of scales to check is only evaluated when the pan operation starts.

### DIFF
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -44,7 +44,7 @@ export default {
     }
     if (Object.prototype.hasOwnProperty.call(options.zoom, 'overScaleMode')
       || Object.prototype.hasOwnProperty.call(options.pan, 'overScaleMode')) {
-      console.warn('The option `overScaleMode` is no longer deprecated. Please use `scaleMode` instead (and update `mode` as desired).');
+      console.warn('The option `overScaleMode` is deprecated. Please use `scaleMode` instead (and update `mode` as desired).');
     }
 
     if (Hammer) {
--- a/src/plugin.js
+++ b/src/plugin.js
@@ -42,6 +42,10 @@ export default {
     if (Object.prototype.hasOwnProperty.call(options.zoom, 'enabled')) {
       console.warn('The option `zoom.enabled` is no longer supported. Please use `zoom.wheel.enabled`, `zoom.drag.enabled`, or `zoom.pinch.enabled`.');
     }
+    if (Object.prototype.hasOwnProperty.call(options.zoom, 'overScaleMode')
+      || Object.prototype.hasOwnProperty.call(options.pan, 'overScaleMode')) {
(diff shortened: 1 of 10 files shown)


src/plugin.js
```diff
@@ -42,6 +42,10 @@ export default {
     if (Object.prototype.hasOwnProperty.call(options.zoom, 'enabled')) {
       console.warn('The option `zoom.enabled` is no longer supported. Please use `zoom.wheel.enabled`, `zoom.drag.enabled`, or `zoom.pinch.enabled`.');
     }
+    if (Object.prototype.hasOwnProperty.call(options.zoom, 'overScaleMode')
+      || Object.prototype.hasOwnProperty.call(options.pan, 'overScaleMode')) {
+      console.warn('The option `overScaleMode` is no longer deprecated. Please use `scaleMode` instead (and update `mode` as desired).');
```
kurkle, line 47: Wording

kurkle (approved):

PR #659, opened 2022-05-20
Export zoomRect and make it more configurable

joshkel:
As discussed in #608, normal zooming supports customizing zoom behavior using zoomFunctions, but the drag-to-zoom functionality that's implemented by zoomRect bypasses zoomFunctions and calls updateRange correctly.

I have a custom scale implementation that relies on zoomFunctions, so trying to enable drag-to-zoom causes problems. (Specifically, I'm breaking up the axis into viewports so that I can stack multiple series within this chart - somewhat like [this example](https://felixfan.github.io/figure2016/stacking1-1.png). The scales for the individual portions of the chart shouldn't be directly zoomed, because their layout is managed by a separate, controlling scale, so their zoomFunctions are no-ops.)

Zooming by rectangle appears to be a distinctly different interface than zooming by percentage and focal point, so this PR adds a zoomRectFunctions to let scales customize that behavior and cover this use case.

I'm also now exposing zoomRect as part of the public API, like zoom and zoomScale.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,4 +1,4 @@
-import { Plugin, ChartType, Chart, Scale, UpdateMode, ScaleTypeRegistry } from 'chart.js';
+import { Plugin, ChartType, Scale, UpdateMode, ScaleTypeRegistry } from 'chart.js';
 import { DistributiveArray } from 'chart.js/types/utils';
 import { LimitOptions, ZoomPluginOptions } from './options';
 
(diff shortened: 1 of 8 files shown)


types/index.d.ts
```diff
@@ -40,15 +42,8 @@ type ScaleFunctions<T> = {
...
 
 export default Zoom;
-
-export function pan(chart: Chart, amount: PanAmount, scales?: Scale[], mode?: UpdateMode): void;
-export function zoom(chart: Chart, amount: ZoomAmount, mode?: UpdateMode): void;
-export function zoomScale(chart: Chart, scaleId: string, range: ScaleRange, mode?: UpdateMode): void;
-export function resetZoom(chart: Chart, mode?: UpdateMode): void;
-export function getZoomLevel(chart: Chart): number;
-export function getInitialScaleBounds(chart: Chart): Record<string, {min: number, max: number}>;
-export function isZoomedOrPanned(chart: Chart): boolean;
```
kurkle, line 54: aren't these available anymore?

joshkel: I made a mistake; I saw that they weren't available in `index.js` (the non-ESM build), but I missed that they are available in `index.esm.js` (the ESM build). And, in some local testing, I was somehow using a non-ESM setup, so the functions weren't available.

I restored these.

kurkle (approved):

PR #672, opened 2022-09-01
Allow canceling drag-to-zoom on Escape key

joshkel:
This makes the behavior of drag-to-zoom closer to the browser's standard drag-and-drop behavior.

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -34,6 +34,18 @@ export function mouseMove(chart, event) {
   }
 }
 
+function keyDown(chart, event) {
+  const state = getState(chart);
+  if (!state.dragStart || event.key !== 'Escape') {
+    return;
+  }
+
+  removeHandler(chart, 'keydown');
+  state.dragging = false;
+  state.dragStart = state.dragEnd = null;
+  chart.update('none');
+}
+
 function zoomStart(chart, event, zoomOptions) {
   const {onZoomStart, onZoomRejected} = zoomOptions;
   if (onZoomStart) {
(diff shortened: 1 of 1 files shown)


kurkle (commented):
Maybe mention this in docs?

stockiNail:
@joshkel do you think it could make sense to have an option (i.e., `cancelKey`) in the drag options where the user can decide which key should be pressed to cancel the dragging, instead of Escape only?

kurkle:
> @joshkel do you think it could make sense to have an option (i.e., `cancelKey`) in the drag options where the user can decide which key should be pressed to cancel the dragging, instead of Escape only?
> 
> EDIT: default is `Escape`.

I think this can be added when another key is requested.

kurkle (approved):

stockiNail:
> > @joshkel do you think it could make sense to have an option (i.e., `cancelKey`) in the drag options where the user can decide which key should be pressed to cancel the dragging, instead of Escape only?
> > EDIT: default is `Escape`.
> 
> I think this can be added when another key is requested.

Ok for me! Probably I will need in my project but not now

cammytang:
Hey @kurkle, what should I do if I want to cancel drag-to-zoom on `shift` key released. I want to handle it similarly to the ESC key keydown scenario. `shift` is the `modifierKey` in my options. See
```
      drag: {
                        enabled: true,
                        modifierKey: 'shift',
                 }
```
I can't get or set the `dragStart` & `dragEnd` as they are inner props.
Could you give some advice? This problem has troubled me for a while. Thanks so much.

helloint:
Maybe, if modifier key is being used to enable the drag-to-zoom, better to keep monitoring this modifier key, when it is released, cancel the zoom immediately.

Issue #690, opened 2022-11-03
Zoom edge cases


AsturaPhoenix:
I ran into a couple edge cases zooming with dynamic data and limits:

* [unexpected `isZoomedOrPanned()` true after `resetZoom` if data changed while zoomed](https://codepen.io/AsturaPhoenix/pen/JjZKoNR)
* [jitter on zoom-in with fully constrained limits (like a "locked" axis)](https://codepen.io/AsturaPhoenix/pen/GRGqRRp)
* [zoom out against a limit vanishes to a no-op as the focal point approaches the other side of the chart](https://codepen.io/AsturaPhoenix/pen/wvXWBvj)

The first issue seems related to, and can be fixed by fixing, `getInitialScaleBounds()` not updating the scale limits (relying on a previous zoom or pan operation to have done so). An easy fix that covers the `resetZoom` case is to call [`storeOriginalScaleLimits` at the beginning of `getInitialScaleBounds`](https://github.com/chartjs/chartjs-plugin-zoom/compare/master...AsturaPhoenix:chartjs-plugin-zoom:reset-zoom?expand=1#diff-8ab41fe13597e1554b5d6b4c227b5f123ff2d6726a7f3688a8b8d1224fe1d4f3), though this does still return stale bounds until `resetZoom` is actually called; a full fix would be more involved, and it's not clear to me that the functionality is exposed on scales that would make for something correct in all cases (e.g. `"original"` limits as data change while zoomed for scales with unset `min`/`max`, especially if they pad).

[...]

PR #708, opened 2022-11-23
Handling zoom edge cases (#690)

AsturaPhoenix:
This PR includes two commits fixing the following edge cases (see #690 for details and demos):

* unexpected `isZoomedOrPanned()` true after `resetZoom` if data changed while zoomed
* jitter on zoom-in with fully constrained limits (like a "locked" axis)
* zoom out against a limit vanishes to a no-op as the focal point approaches the other side of the chart

One remaining concern is that this can affect zooming out from category scales in a potentially unintuitive way; further adjustments may be possible once we decide what the desired behavior is.

Thank you; let me know if you'd like any edits!

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -199,6 +199,7 @@ export function pan(chart, delta, enabledScales, transition = 'none') {
 
 export function getInitialScaleBounds(chart) {
   const state = getState(chart);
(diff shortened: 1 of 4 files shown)


kurkle (approved):
LGTM

etimberg (approved):

PR #732, opened 2023-02-19
Fix drag rectangle calculation when inside shadow BOM / WebComponent

### LINKED ISSUE #726: Chart / Zoom inside a webcomponent is offset
"chart.js": "^4.2.0"
"chartjs-plugin-zoom": "^2.0.0"

Hi,

I try to debug something, I hope to find someone having the same issue. It's really hard to debug.

I use ChartJs/Zoom inside CustomElement in Vue.js js 3.
When I zoom in a web component, the zoom is offset a few pixels (30px). 

When I use the Chart in Storybook without a webcomponent everything seem fine.

I think it's really related to the WebComponent
Something with the shadowDom ?

In handler.js, I'am playing with mouseDown(chart, event), mouseUpchart, event) and getState(chart), and computeDragRect trying to understand the bug.

Maybe the mouseEvent is related to webcomponent/shadowDom, not the canvas?

Will try to debug this today, and give more details, but any idea would be appreciated

======
EDIT:

When we click on a webcomponent, the click event is on the shadowDom Element, not the canvas element. So the click position is wrong (because of padding and margin surrounding the canvas).

=====
EDIT 2 :

See https://stackoverflow.com/questions/57963312/get-event-target-inside-a-web-component

Here https://github.com/chartjs/chartjs-plugin-zoom/blob/master/src/handlers.js#L85
```
export function computeDragRect(chart, mode, beginPoint, endPoint) {
  const {left: offsetX, top: offsetY} = beginPoint.target.getBoundingClientRect();
```
The bug is here : target under a webcomponent is the shadowRoot, not the canvas

https://user-images.githubusercontent.com/39559391/217631216-bebb04df-2df9-413d-a8b1-229bd4671164.mov
test

https://user-images.githubusercontent.com/39559391/217631942-aa839290-5c92-4f64-849b-6c8111785f9a.mov

CodingMarco:
Fixes #726.

The edges of the drag rectangle were calculated incorrectly when the chart is placed inside a shadow BOM (eg. inside a web component). Chart.js itself already has fixes for that (see [Chart.js PRs](https://github.com/chartjs/Chart.js/pulls?q=is%3Apr+shadow+dom)). Its getRelativePosition() function always returns the correct coordinates.
So import and use this function from Chart.js.

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -1,6 +1,6 @@
 import {directionEnabled, debounce, keyNotPressed, getModifierKey, keyPressed} from './utils';
 import {zoom, zoomRect} from './core';
-import {callback as call} from 'chart.js/helpers';
+import {callback as call, getRelativePosition} from 'chart.js/helpers';
 import {getState} from './state';
 
 function removeHandler(chart, type) {
@@ -49,11 +49,7 @@ function keyDown(chart, event) {
 function zoomStart(chart, event, zoomOptions) {
   const {onZoomStart, onZoomRejected} = zoomOptions;
   if (onZoomStart) {
-    const {left: offsetX, top: offsetY} = event.target.getBoundingClientRect();
-    const point = {
-      x: event.clientX - offsetX,
-      y: event.clientY - offsetY
-    };
+    const point = getRelativePosition(event, chart);
     if (call(onZoomStart, [{chart, event, point}]) === false) {
       call(onZoomRejected, [{chart, event}]);
       return false;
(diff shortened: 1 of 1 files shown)


hugo-valcourt (approved):

kurkle (approved):

kurkle:
CI fails because of FF, fix is same as in Chart.js: https://github.com/chartjs/Chart.js/pull/11165/files
Feel free to create another PR for that fix

jcphill:
getRelativePosition seems to give the wrong position when mouseup occurs over a different element (as when a drag ends outside the chart canvas), which is causing the regression in https://github.com/chartjs/chartjs-plugin-zoom/issues/770

Issue #735, opened 2023-02-24
UpdateModeEnum - enum collision when using chartjs@4.2.0 & chartjs-plugin-zoom@2.0.0


infrat:
Hello.

I have an angular project where I'm using `chart.js@4.2.0` and I'm trying to integrate `chartjs-plugin-zoom@2.0.0`.
After importing `chartjs-plugin-zoom` into component where I'm setting up instance of `chart.js`, I'm getting error as follows:

```
Error: node_modules/chart.js/dist/types/index.d.ts:570:21 - error TS2567: Enum declarations can only merge with namespace or other enum declarations.

570 export declare enum UpdateModeEnum {
                        ~~~~~~~~~~~~~~

  node_modules/chartjs-plugin-zoom/types/index.d.ts:16:14
    16   const enum UpdateModeEnum {
                    ~~~~~~~~~~~~~~
    'UpdateModeEnum' was also declared here.


Error: node_modules/chartjs-plugin-zoom/types/index.d.ts:16:14 - error TS2567: Enum declarations can only merge with namespace or other enum declarations.

16   const enum UpdateModeEnum {
                ~~~~~~~~~~~~~~

  node_modules/chart.js/dist/types/index.d.ts:570:21
    570 export declare enum UpdateModeEnum {
                            ~~~~~~~~~~~~~~
    'UpdateModeEnum' was also declared here.




✖ Failed to compile.
```

Changing `const enum` to `enum` in [`node_modules/chartjs-plugin-zoom/types/index.d.ts@16`](https://github.com/chartjs/chartjs-plugin-zoom/blob/master/types/index.d.ts#L16) makes the compilation successful in my case. It seems that this change has been introduced in this PR: https://github.com/chartjs/chartjs-plugin-zoom/pull/695/files#diff-093ad82a25aee498b11febf1cdcb6546e4d223ffcb49ed69cc275ac27ce0ccceL16

All the dependencies in my project I have in my `package.json`:

```
"dependencies": {
    "@angular/animations": "^15.1.5",
    "@angular/common": "^15.1.5",
    "@angular/compiler": "^15.1.5",
    "@angular/core": "^15.1.5",
    "@angular/forms": "^15.1.5",
    "@angular/localize": "^15.1.5",
    "@angular/platform-browser": "^15.1.5",
    "@angular/platform-browser-dynamic": "^15.1.5",
    "@angular/router": "^15.1.5",
[...]

Issue #731, opened 2023-02-16
TypeScript error since updating chart.js to v4.2.1


StefanNedelchev:
After updating chart.js from 4.2.0 to 4.2.1 I started getting this TypeScrip error at build:
![image](https://user-images.githubusercontent.com/15238282/219469823-515c8bbc-eb03-4f00-8202-bd8e408cb891.png)

I use v2.0.0 of the zoom plugin. Reverting chart.js back to 4.2.0 solved the problem

dt-bib:
Same issue and revert to 4.2.0 solve the problem

Related to this issue : https://github.com/chartjs/chartjs-plugin-zoom/issues/721

Issue #721, opened 2023-01-16
Error TS2567 when using this plugin in version 2.0.0


hemmoleg:
Hi,

when I install this plugin and import it I get two errors:

```
Error: node_modules/chart.js/types/index.esm.d.ts:561:21 - error TS2567: Enum declarations can only merge with namespace or other enum declarations.

561 export declare enum UpdateModeEnum {
                        ~~~~~~~~~~~~~~

  node_modules/chartjs-plugin-zoom/types/index.d.ts:16:14
    16   const enum UpdateModeEnum {
                    ~~~~~~~~~~~~~~
    'UpdateModeEnum' was also declared here.
```
    
and

```
Error: node_modules/chartjs-plugin-zoom/types/index.d.ts:16:14 - error TS2567: Enum declarations can only merge with namespace or other enum declarations.

16   const enum UpdateModeEnum {
                ~~~~~~~~~~~~~~

  node_modules/chart.js/types/index.esm.d.ts:561:21
    561 export declare enum UpdateModeEnum {
                            ~~~~~~~~~~~~~~
    'UpdateModeEnum' was also declared here.
```
    
I am using chartjs 3.8.0.
When using version 1.2.1 of this plugin everything works. However I dont think thats a proper solution.
Has anyone else run into this issue?

krisanselmo:
Hi. I have the same issue with chartjs 3.9.1.
Thx for the temporary solution.

danielmcquillen:
Same issue.

CoooWeee:
any updates?

elinake:
Issue is replicated here: https://stackblitz.com/edit/angular-chart-js-a2tqkw?file=node_modules/chartjs-plugin-zoom/types/index.d.ts

When you click Zoom in Chart.register, you will see that UpdateModeEnum in chartjs-plugin-zoom has red squiggly line with error:
```
Enum declarations can only merge with namespace or other enum declarations.(2567)
index.d.ts(570, 21): 'UpdateModeEnum' was also declared here.
```
This project does build, however, in real projects it is often not even allowed to disable typescript entirely.

yKicchan:
I have the same issue with chart.js 4.2.1 and typescript 4.4.4 too.

PR #695, opened 2022-11-17
add-v4-compatibility

LeeLenaleee:
Resolves https://github.com/chartjs/chartjs-plugin-zoom/issues/694

PR #736, opened 2023-02-27
Modify UpdateModeEnum to not be const

SebastiaanSafeguard:
Closes https://github.com/chartjs/chartjs-plugin-zoom/issues/735
Closes https://github.com/chartjs/chartjs-plugin-zoom/issues/721
Closes https://github.com/chartjs/chartjs-plugin-zoom/issues/731

This enum was changed to a normal enum instead of using const in https://github.com/chartjs/Chart.js/pull/11095, so the type is no longer matching with Chart.js.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "@typescript-eslint/eslint-plugin": "^5.4.0",
     "@typescript-eslint/parser": "^5.4.0",
     "babel-loader": "^8.3.0",
-    "chart.js": "^4.0.1",
+    "chart.js": "^4.2.1",
     "chartjs-adapter-date-fns": "^2.0.1",
     "chartjs-test-utils": "^0.3.0",
     "concurrently": "^6.0.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "@typescript-eslint/eslint-plugin": "^5.4.0",
         "@typescript-eslint/parser": "^5.4.0",
         "babel-loader": "^8.3.0",
-        "chart.js": "^4.0.1",
+        "chart.js": "^4.2.1",
         "chartjs-adapter-date-fns": "^2.0.1",
         "chartjs-test-utils": "^0.3.0",
         "concurrently": "^6.0.2",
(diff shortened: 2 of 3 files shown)


flogoe (dismissed):

SebastiaanSafeguard:
The change in Chart.js was technically breaking, since the Typescript types in any plugin could not have matched anymore. So bundled with the latest Chart.js version I wouldn't consider this a breaking change. With this change the code does however no longer work with any previous version of Chart.js except for those _before_ the addition of the enum.

elinake:
What are the exact versions of Chart.js that should work before and after the change?

flogoe:
I got it running with the following versions: 

```
...
"chart.js": "4.1.1",
"chartjs-plugin-zoom": "^2.0.0",
...
```

SebastiaanSafeguard:
I would say that based on https://github.com/chartjs/chartjs-plugin-zoom/pull/695/files#diff-093ad82a25aee498b11febf1cdcb6546e4d223ffcb49ed69cc275ac27ce0ccce

2.0.0 of this library introduced `const`for chart.js 4.0.0. Then with 4.2.1`const` was removed in chart.js, so the changes in this PR should make the library compatible with >=4.2.1 and possibly with 3.x depending on the other changes.

kurkle:
CI fails because of FF, fix is same as in Chart.js: https://github.com/chartjs/Chart.js/pull/11165/files
Feel free to create another PR for that fix

package.json
```diff
@@ -79,7 +79,7 @@
     "vuepress-theme-chartjs": "^0.2.0"
   },
   "peerDependencies": {
-    "chart.js": ">=3.2.0"
```
LeeLenaleee, line 82: Why is this necasarry, looks like this should still work fine with v3, otherwise this will be a breaking change and needs a major version bump

kurkle: Let's revert this change, and deal with possible errors with Chart.js v3 in case we encounter them.

kurkle (changes requested):

kurkle (approved):

LeeLenaleee (approved):

PR #737, opened 2023-03-02
Disable canvas acceleration in Firefox 110 in test cases

SebastiaanSafeguard:
See https://github.com/chartjs/Chart.js/pull/11165


LeeLenaleee (approved):

SebastiaanSafeguard:
@kurkle Would you maybe have time to review this? :)

kurkle (approved):

SebastiaanSafeguard:
The result is the timeout, that's not the issue itself. There is always a reason for a timeout to happen.

kurkle:
> The result is the timeout, that's not the issue itself. There is always a reason for a timeout to happen.

Windows workers time out randomly but consistently :)

PR #774, opened 2023-07-25
Fix "original" scale limits with nonlinear pan

joshkel:
Nonlinear scales failed to resolve "original" within panNumericalScale.

To fix this:

* Resolving "original" within panNumericalScale seemed redundant, when updateRange was already doing it, so I instead moved the logic into updateRange and control it with a new special value for the `zoom` parameter.  (Hopefully the use of a special string like this is acceptable as long as JSDoc calls it out - if I should use another approach, please let me know.)
* This change put updateRange over the configured ESLint complexity limit, so I extracted a new getScaleLimits function to reduce complexity.

Add unit tests, both for this bug and for the preexisting but untested    "original" feature.

I wanted JSDoc + TypeScript checks to help validate and document these changes, but that exposed apparent limitations of Chart.js core's type definitions. I'll open a separate PR for those.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,7 +19,7 @@
         "@typescript-eslint/eslint-plugin": "^5.4.0",
         "@typescript-eslint/parser": "^5.4.0",
         "babel-loader": "^8.3.0",
-        "chart.js": "^4.2.1",
+        "chart.js": "^4.3.2",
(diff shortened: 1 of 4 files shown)


kurkle (approved):

trullock:

PR #780, opened 2023-09-04
Add a function for getting zoom range

joshkel:
This adds a function, `getZoomedScaleBounds`, to act as a counterpart to `getInitialScaleBounds` (getting the scale ranges after zoom rather than before) and `zoomScale` (reading, rather than writing, each scale's zoom).

A question for code review: Is it okay to return the internal object directly, or should I return a new object to ensure that it can't be mutated?


trullock:

kurkle:
@joshkel would you be able to resolve the merge conflicts in this?

kurkle (approved):

joshkel:
Thanks for taking care of this, @kurkle. I apologize for not following up.

PR #808, opened 2024-03-09
chore: update rollup

bavoco:

PR #809, opened 2024-03-11
refactor: VuePress to VitePress

bavoco:
This reduces the number of dev dependencies by about 700.

Towards #812


NotTsunami:
By nature, the move from VuePress -> VitePress should also allow us to drop the babel dependencies too, right? A very _brief_ peek at the source suggests it is not used outside of VuePress.

NotTsunami:
I tested it on your branch and `npm run docs` passes with the babel dependencies removed. With `@babel/core`, `@babel/preset-env`, and `babel-loader` removed, another 155 packages are removed. `npm run test` passes with 76/76 tests while `npm run lint` and `npm run build` complete without error.

NotTsunami:
You should be able to remove `ng-hammerjs` as well, since it was only used in documentation: https://github.com/chartjs/chartjs-plugin-zoom/blob/bfd207d9108cbcd612b8f4c2b0221ef8e7ed373b/docs/.vuepress/config.js#L89-L90

bavoco:
Great! If #808 gets merged there won't be two versions of Rollup and related plugins further reducing the count.

kurkle:
> You should be able to remove `ng-hammerjs` as well, since it was only used in VuePress:
> 
> https://github.com/chartjs/chartjs-plugin-zoom/blob/bfd207d9108cbcd612b8f4c2b0221ef8e7ed373b/docs/.vuepress/config.js#L89-L90

Hammer is needed by the zoom plugin, and ng-hammerjs can be used as replacemend (it does not require window).
So I don't think that can be removed (until this plugin gets rid of the dependency)

kurkle:
I did not have time to take a look at this yet, but I'm game getting rid of VuePress.

NotTsunami:
> > You should be able to remove `ng-hammerjs` as well, since it was only used in VuePress:
> > https://github.com/chartjs/chartjs-plugin-zoom/blob/bfd207d9108cbcd612b8f4c2b0221ef8e7ed373b/docs/.vuepress/config.js#L89-L90
> 
> Hammer is needed by the zoom plugin, and ng-hammerjs can be used as replacemend (it does not require window). So I don't think that can be removed (until this plugin gets rid of the dependency)

Yes, hammer.js is still needed, but ng-hammerjs was only used for VuePress, as it appeared, so I was suggesting the move to VitePress could remove ng-hammerjs. I'm not suggesting to remove hammer entirely.

bavoco:
Good to know there is interest in this! I will have to take another look at it, because some things have changed in the meantime (VitePress 1.0, ...).

stockiNail:
> I did not have time to take a look at this yet, but I'm game getting rid of VuePress.

@kurkle does it mean we will migrate the other plugins (i.e. annotation) too? I personally agree with that because many of security alerts are coming from a Vuepress version.

kurkle:
> @kurkle does it mean we will migrate the other plugins (i.e. annotation) too? I personally agree with that because many of security alerts are coming from a Vuepress version.

Hopefully 😅

stockiNail:
@kurkle fine! I'll open a task in the annotation as soon as I can ;)

kurkle:
> What’s the consensus on moving the doc portion to a separate package.json like the main chart.js repo does?

I'm not against that. Would help separating the dependencies. Probably should change to pnpm too.

bavoco:
This is ready for an initial review, please have a look if you like the approach. If you like it I'll update the docs with the changes from the past weeks (the content moved to new files, so it's a fairly manual process).

For some reason I cannot lint locally (windows), it hangs indefinitely. Anyone else on windows?

kurkle:
> This is ready for an initial review, please have a look if you like the approach. If you like it I'll update the docs with the changes from the past weeks (the content moved to new files, so it's a fairly manual process).
> 
> For some reason I cannot lint locally (windows), it hangs indefinitely. Anyone else on windows?

Took a quick first look and yes, I like the approach.

Some issues I'd like resolved are:
 * The canvas should have a container (div) in samples
 * Not sure why in the samples the chart is initially draw with size 0,0. I think that is the reason animations are ugly.
 * Need to verify version linking works
 * Pan-region sample is 404
 
For the lint issue, I'd assume its the vitepress cache folder that causes trouple, you could try adding it to .eslintignore

Issue #631, opened 2022-03-11
onZoomStart is not called when using zoom via pinch on mobile.


airwalk741:
onZoomStart is not called when using zoom via pinch on mobile_web.

OnZoomStart is called when zooming with the mouse wheel, but not when using pinch on mobile_web.

https://codesandbox.io/s/vue-chart-3-chart-js-issue-template-forked-ce6dtz?file=/src/App.vue

("zoom start" warning window does not appear)

Isaac96SM:
I'm also having this issue. Is it planned to be solved soon?

kurkle:
Fixed by #810

Issue #790, opened 2023-11-01
onZoomStart is not called for pinch


tomeronen:
if pinch is enabled, then zoom happens without the onZoomStart being called. I think this is a quick fix. Can i open a PR to add the onZoomStart function before the zoom call? 
```
function handlePinch(chart, state, e) {
  if (state.scale) {
    const {center, pointers} = e;
    // Hammer reports the total scaling. We need the incremental amount
    const zoomPercent = 1 / state.scale * e.scale;
    const rect = e.target.getBoundingClientRect();
    const pinch = pinchAxes(pointers[0], pointers[1]);
    const mode = state.options.zoom.mode;
    const amount = {
      x: pinch.x && directionEnabled(mode, 'x', chart) ? zoomPercent : 1,
      y: pinch.y && directionEnabled(mode, 'y', chart) ? zoomPercent : 1,
      focalPoint: {
        x: center.x - rect.left,
        y: center.y - rect.top
      }
    };

    zoom(chart, amount);

    // Keep track of overall scale
    state.scale = e.scale;
  }
}
```

kurkle:
Fixed by #810

romor:
When returning `false` in the onZoomStart() callback, the zoom operation should be cancelled according documentation.

However, the pinch event is not cancelled for me and the zoom is done although it should be aborted.

Thus, I think this ticket shall not be closed as it is not completely finished.

PR #810, opened 2024-03-15
fix: trigger `onZoomStart` from `pinchStart`

ZaLiTHkA:
this simply ensures that the custom `onZoomStart` handler is also executed when zooming is initiated with "pinch".

there seems to be a fair difference in the logic structure between the `handlers.js` and `hammer.js` files, so I tried to keep this in line with what I already saw in `hammer.js` directly.

I did not add any tests for this, though I feel like we probably should.. perhaps one of the repo maintainers could help determine where this test should go? I don't see any others that cover the "pinch" functionality specifically.

this follows on from PR #487, and fixes issues #631 and #790.

### DIFF
--- a/src/hammer.js
+++ b/src/hammer.js
@@ -68,6 +68,7 @@ function handlePinch(chart, state, e) {
 
 function startPinch(chart, state) {
   if (state.options.zoom.pinch.enabled) {
+    call(state.options.zoom.onZoomStart, [{chart}]);
     state.scale = 1;
   }
 }


LeeLenaleee (approved):

trullock:

ZaLiTHkA:
thanks @trullock, I'll try it out. I see your fork consolidates a fair amount of work from multiple PRs.. just out of curiosity, are you actively using it in one of your own projects as well?

I still think it would be nice to try get these changes into this root project, if possible.. this way the fixes and improvements could reach more users more easily, including your contributions.

so I'll leave this PR open for a little bit longer, just to see where this all goes. 👍🏼

trullock:
Absolutely, you should leave this open.

I dont want to be maintaining a fork 🙈

ZaLiTHkA:
> I dont want to be maintaining a fork 🙈

yeah, completely understand that. it's definitely not for just anybody.. I only wish I had the time to even consider making commitments like that.

trullock:
My fork works for me, thats all I'm guaranteeing! 😂

kurkle (approved):

ZaLiTHkA:
happy days! thanks @kurkle, looking forward to switching my `package.json` back to the official package. 💯

Issue #807, opened 2024-03-06
Chartjs-plugin-zoom drag issue


bjlee3075:
Hi
I am drawing a graph using chart.js. Below is my code.
zoom works fine. However, there is a little problem.
If the mouse leaves the chart while dragging, it will not zoom in properly.
Like the first picture, the starting point of the drag is 2023, and the end of the drag is outside the chart, not inside the chart.
The starting point of drag is 2023, but as shown in the second picture, 2011 became the starting point.
Is there any way to prevent this?

Thanks,

![2](https://github.com/chartjs/chartjs-plugin-zoom/assets/143692239/5985ea6b-c404-4fac-8254-f722914990d9)
![1](https://github.com/chartjs/chartjs-plugin-zoom/assets/143692239/661c3dc7-ae74-48a9-adfe-0980708f9bb5)

			zoom: {
				pan: {
					enabled: false,
					mode: 'x',
				},
				zoom: {
					wheel: {
						enabled: false,
					},
					pinch: {
						enabled: false,
					},
					drag: {
						enabled: true,
						mode: 'x',
						borderColor: 'rgba(0, 0, 0, 0.5)', 
						borderWidth: 1, 
						backgroundColor: 'rgba(0, 0, 0, 0.3)',
					},
					mode: 'x',
					
					onZoomComplete: function (context) {
						const chart = context.chart;
						const xScale = chart.scales.x;
						const displayedRange = xScale.max - xScale.min;
						const oneYear = 1000 * 60 * 60 * 24 * 365;
						isLongRange = displayedRange >= oneYear * 10;

						chart.options.scales.x.time.unit = 'month';

						if (displayedRange < oneYear * 3) {
							chart.options.scales.x.ticks.stepSize = 1;
						} else if (displayedRange < oneYear * 6) {
							chart.options.scales.x.ticks.stepSize = 2;
						} else if (displayedRange < oneYear * 10) {
							chart.options.scales.x.ticks.stepSize = 6;
						} else {
							chart.options.scales.x.time.unit = 'year';
							delete chart.options.scales.x.ticks.stepSize; 
						}

						chart.update();
					},
				},
				limits: { 
					x: {
						min: new Date(originalXMin).getTime(), 
						max: new Date(originalXMax).getTime() 
					}
				},
			},

Issue #770, opened 2023-06-19
Drag to the outside of the chart doesn't work correctly


lukicenturi:
Drag to the outside of the chart doesn't work correctly
It's only happen in `2.0.1`. I downgrade to `2.0.0`, it works fine.

![image](https://github.com/chartjs/chartjs-plugin-zoom/assets/26648140/23aacd6b-5f0a-42a3-8d81-c7a764b3199c)
In `2.0.1`, it seems that it calculated the length between `mousedown` and `mouseup`, so it shows longer range than expected.
(I expect the range starts on the `mousedown` point, till the end of the chart)

jcphill:
I see this same issue (with chart.js 3.9.1 and react 17.0.2) and can confirm that downgrading from 2.0.1 to 2.0.0 fixes it.

I believe the cause is that getRelativePosition (called in computeDragRect, which is called in mouseUp) is returning a position relative to whatever element the mouse was over when the endPointEvent mouseup happened rather than the position relative to the chart canvas.

https://github.com/chartjs/chartjs-plugin-zoom/blob/bfd207d9108cbcd612b8f4c2b0221ef8e7ed373b/src/handlers.js#L80-L120

sqlAlchemyNoob2004:
I was just about to create an issue for this. Unsure if there is active development on this plugin, but what I think you should follow the `chartjs` core implementation of `getRelativePosition` and write your own.

The reason theirs doesn't work for you is because they seem to assume that any events passed to that method have an equivalent `target` and `currentTarget`, i.e. they all originate from the canvas. This isn't the case for the zoom plugin because the `mouseUp` listener for drag detection is attached to the `canvas.ownerDocument`. 

This gets you into situations where, when dragging out of the canvas, your start and end points get flipped. For example:
* I have a 500px height canvas 
* I'm trying to select from the middle (250), to the end (500)
* I drag from the middle to one pixel outside of the canvas, 1px into some sibling element
* `e.target` is now that sibling element, so `offsetY` becomes 1
* I now have `start.y = 250` and `end.y = 1` which just gets flipped further down the call chain
* on mouseUp I've selected selected the first half of the chart (1-250) instead of the last half (250-500) like I intended.

Another method to get canvas position would be to, instead of relying on `offsetY/X`, use `canvas.getBoundingClientRect()`. Then just subtract `clientX/Y` by `left/top`.

[...]

PR #811, opened 2024-03-17
Fix zoom outside of chart area

gopal-panigrahi:
Fix: #770
Fix: #807  

Changes:

1. Attaches `mousemove` event to `canvas.ownerDocument` for calculating drag rectangle even when cursor is outside the chart area.
2. checks if endpoint of drag rectangle is outside the chart area then endpoint is calculated based on `clientX/Y - chartArea.left/top`
3. clips drag rectangle to the chart area.


gopal-panigrahi:
Hi @kurkle,
Please can you review the PR.

zloveless:
Hi @gopal-panigrahi 
I believe this would fix an issue we're having in our charts. Is there any portable solution I can try on my end to see if it works?

gopal-panigrahi:
In our project, I had to create a patch file for these changes, you can try the same.
I used the [patch-package](https://www.npmjs.com/package/patch-package) library.

zloveless:
> Hi @zloveless, In our project, I had to create a patch file for these changes, you can try the same. I used the [patch-package](https://www.npmjs.com/package/patch-package) library.

Thanks! I found some code in one of the tracking issues on this here... I just used added a callback to onZoomStart() to check if the zoom action is in the chart area. I still wish one of these PR's that fix it would get merged.

joshkel:
Thanks for working on this, @gopal-panigrahi. I'm curious about the changes in d9fd526d225f7b5b8e335662b386719301a52f96 - to me, the behavior feels cleaner without this change (letting the drag rectangle pin to the edges of the chart area, instead of the edges of the canvas). Were there specific issues that you ran into, or is it more of a UX preference?

gopal-panigrahi:
Hi @joshkel ,
I had bounded the drag rectangle to pin to the edges of the chart area, instead of the edges of the canvas but then zoom is enabled for x and y axis region, which causes issues while creating rectangles.

gopal-panigrahi:
Hi @kurkle ,
Are there any comments on this change?. 
It fixes the issue of drag rectangle not updating when the cursor moved out of the chart.
Do you have an opinion on the boundary of the drag rectangle ?

kurkle:
> Hi @kurkle , Are there any comments on this change?. It fixes the issue of drag rectangle not updating when the cursor moved out of the chart. Do you have an opinion on the boundary of the drag rectangle ?

Hi @gopal-panigrahi, and than you for the PR.

I'm not sure if the ownerDocument works for all use cases, maybe there should be an option to enable this?
For the boundary, intuitively I'd think it should be chartArea. But not sure about that either. 
Why did you go for canvas area?

gopal-panigrahi:
Hi @kurkle ,
1. I used ownerDocument for mousemove, because the mouseup event was also attached to ownerDocument, and seemed to work fine.
2. Chose canvasArea instead of chartArea because the zoom trigger is allowed outside the chartArea also, like on the axis, and that creates issues with drag rectangle calculations

With boundary fixed to chartArea:  https://jsbin.com/sogeqanofa/1/edit?js,output, Try creating drag rectangle at across the axis to chartArea

With boundary fixed to canvasArea: https://jsbin.com/napadidebu/1/edit?html,js,output

kurkle:
> Hi @kurkle ,
> 
> 1. I used ownerDocument for mousemove, because the mouseup event was also attached to ownerDocument, and seemed to work fine.
> 2. Chose canvasArea instead of chartArea because the zoom trigger is allowed outside the chartArea also, like on the axis, and that creates issues with drag rectangle calculations
> 
> With boundary fixed to chartArea: https://jsbin.com/doxidebene/1/edit?js,output, Try creating drag rectangle at across the axis to chartArea
> 
> With boundary fixed to canvasArea: https://jsbin.com/vunezavomu/1/edit?js,output

Thanks!
Looks like the getPointPosition function is not really needed, because the added min/max limits do the job already. Am I missing something?

gopal-panigrahi:
I had written getPointPosition function specifically for endPoint outside of canvas area. 
When we use getRelativePosition for points outside chart area, it mis-calculates the end point position, i changed the calculation for points outside the canvas area.

code sample : https://jsbin.com/culesidobe/1/edit?html,output

[mis-calculation.webm](https://github.com/user-attachments/assets/f02169f4-7c95-40da-a1ad-94758101b396)

kurkle (approved):

PR #823, opened 2024-07-29
Add isZoomingOrPanning method

joshkel:
I'd like to ensure that browser click events don't trigger if the user is zooming or panning (similar to how Chart.js prevents its own click events from triggering if the user is zooming or panning). Adding a new `isZoomingOrPanning` method seems like the easiest solution.

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -225,3 +225,8 @@ export function isZoomedOrPanned(chart) {
 
   return false;
 }
+
+export function isZoomingOrPanning(chart) {
+  const state = getState(chart);
+  return state.panning || state.dragging;
+}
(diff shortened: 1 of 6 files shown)


src/core.js
```diff
@@ -225,3 +225,8 @@ export function isZoomedOrPanned(chart) {
 
   return false;
 }
+
+export function isZoomingOrPanning(chart) {
```
LeeLenaleee, line 229: Shouldn't this be called `IsPanningOrDragging`?

kurkle (approved):

src/core.js
kurkle: @LeeLenaleee to me the reasoning sound good, do you have any strong opinions about the name?

LeeLenaleee (approved):

PR #836, opened 2024-10-01
fixes getZoomLevel for hidden scales

trullock:
Hidden scales aren't zoomed, this confuses `getZoomLevel()`


kurkle:
@trullock are you sure hidden scales are not zoomed? What if all the scales are hidden on a chart, how does one zoom that?

Issue #768, opened 2023-06-12
Feature Request: add ability to limit upper and lower bounds of min an max scale values independently


10man:
Hello,

First off, thanks for the plugin, super awesome. :tophat: 
I'm using the latest (as of 06-11-23) from [cdn.jsdelivr](https://cdn.jsdelivr.net/npm/chartjs-plugin-zoom) for both chart.js and the plugin.

The idea here is to fix (or limit) an axis value and have the zoom affect the data scale in a separately controllable fashion. For example; say we have a dataset that extends away from 0 and we want to zoom keeping 0 in view.

Currently we have:
``` javascript
limits: {
    y: {
        min: 'original'|value,
        max: 'original'|value',
        minRange: value
    }
}
```
I purpose the following:
``` javascript
limits: {
    y: {
        min: 'original'|value,
        max: 'original'|value,
        minRange: value,
        maxMin: 'original'|value,
        minMax: 'original'|value
    }
}
```
Where:
`minMax` sets the *maximum* value that the `min` scale could take under zoom.
`maxMin` sets the *minimum* value that the `max` scale could take under zoom.
Setting to `original` would have the standard implications. Now, if both `min` and `minMax` were set to `original`, that would disable changes to that scale value under zoom.

10man:
I wouldn't mind doing the work if someone can point me in the right direction. That is, if anyone but me thinks it would be useful...

I took a brief look but I don't know this codebase so I wasn't sure where the bast place to start was. I would offer to do the `maxRange` https://github.com/chartjs/chartjs-plugin-zoom/issues/756 update as well if that would help. Seems like that would be in the same area...

estanglerbm:
Could you show an example?  (I can keep 0 in view by doing a drag zoom, and I can't imagine other examples.)

kurkle:
To me, the proposal (below) seems a bit confusing due to the names.

```js
limits: {
    y: {
        min: 'original'|value,
        max: 'original'|value,
        minRange: value,
        maxMin: 'original'|value,
        minMax: 'original'|value
    }
}
```

How about this instead (with some extensions):
```js
limits: {
    y: {
        min: 'original' | 'locked' | value | { min: value, max: value },
        max: 'original' | 'locked' | value | { min: value, max: value },
        minRange: value,
        maxRange: value,
    }
}
```

PR #889, opened 2024-11-17
fix: trigger onZoom from zoomScale

### LINKED ISSUE #775: onZoom callbacks for zoomScale function
When using `zoomScale`, callbacks for zoom (onZoom etc.) are not triggered.

kurkle:
fix #775

### DIFF
--- a/src/core.js
+++ b/src/core.js
@@ -108,10 +108,13 @@ export function zoomRect(chart, p0, p1, transition = 'none') {
 }
 
 export function zoomScale(chart, scaleId, range, transition = 'none') {
-  storeOriginalScaleLimits(chart, getState(chart));
+  const state = getState(chart);
+  storeOriginalScaleLimits(chart, state);
   const scale = chart.scales[scaleId];
   updateRange(scale, range, undefined, true);
   chart.update(transition);
+
+  call(state.options.zoom?.onZoom, [{chart}]);
 }
 
 export function resetZoom(chart, transition = 'default') {
@@ -130,6 +133,7 @@ export function resetZoom(chart, transition = 'default') {
     delete state.updatedScaleLimits[scale.id];
   });
   chart.update(transition);
+
   call(state.options.zoom.onZoomComplete, [{chart}]);
 }
 
--- a/test/specs/api.spec.js
+++ b/test/specs/api.spec.js
@@ -240,6 +240,35 @@ describe('api', function() {
     });
   });
 
+  describe('zoomScale', function() {
+    it('should call onZoom', function() {
+      const zoomSpy = jasmine.createSpy('start');
+      const chart = window.acquireChart({
+        type: 'scatter',
+        data: {
+          datasets: [
+            {
+              data: [{x: 1, y: 1}, {x: 10, y: 10}],
+            },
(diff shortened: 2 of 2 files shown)


yubiuser:
This update broke our use case and I don't know how to mitigate the new behavior. We basically have the same situation as described as in https://github.com/chartjs/chartjs-plugin-zoom/issues/768:

We have a time series bar chart with only positive y-values. When zooming, y-axis should stay fixed at zero, because it does not make sense to lose the zero anchor. We basically used `onZoom` to trigger `chart.zoomScale` to set the new scale with fixed y min to zero.

Screenshot from 2024-11-19 17-40-09


But with this change, this causes an endless loop. Any idea how to circumvent this?

Old code

```
      onZoom: function ({ chart }) {
        // The first time the chart is zoomed, save the maximum initial scale bound
        if (!chart.absMax) chart.absMax = chart.getInitialScaleBounds().y.max;
        // Calculate the maximum value to be shown for the current zoom level
        const zoomMax = chart.absMax / chart.getZoomLevel();
        // Update the y axis scale
        chart.zoomScale("y", { min: 0, max: zoomMax }, "default");
        // Update the y axis ticks and round values to natural numbers
        chart.options.scales.y.ticks.callback = function (value) {
          return value.toFixed(0);
        };
```

kurkle:
Instead of calling zoomScale from onZoom, just specify `min: 0` in the y-scale options?

yubiuser:
> Instead of calling zoomScale from onZoom, just specify min: 0 in the y-scale options?

This does not work, as it specifies the minimum, but not the maximal minimum (it doesn't get negative).
___
```
chart.options.scales.y.min = 0;
chart.options.scales.y.max = chart.absMax / chart.getZoomLevel(); // probably not needed
chart.update();
```

This works, but gives an unpleasant scale jump
Peek 2024-11-20 07-49

Before it was smooth
Peek 2024-11-20 07-55

kurkle:
maybe `chart.update('none')` produces similar result?

yubiuser:
Yes, but now I can only zoom in, and not zoom out again.

```
      onZoom: function ({ chart }) {
        chart.options.scales.y.min = 0;
        chart.update('none');
        }
```

kurkle:
I've created #901 for letting you do something like:

```js
      onZoom: function ({ chart, trigger }) {
        if (trigger === 'api') {
          // Ignore onZoom triggered by the chart.zoomScale api call below
          return;
        }
        // The first time the chart is zoomed, save the maximum initial scale bound
        if (!chart.absMax) chart.absMax = chart.getInitialScaleBounds().y.max;
        // Calculate the maximum value to be shown for the current zoom level
        const zoomMax = chart.absMax / chart.getZoomLevel();
        // Update the y axis scale
        chart.zoomScale("y", { min: 0, max: zoomMax }, "default");
        // Update the y axis ticks and round values to natural numbers
        chart.options.scales.y.ticks.callback = function (value) {
          return value.toFixed(0);
        };
```

yubiuser:
Perfect. Tested it locally compiled and it works like charm.

Peek 2024-11-23 09-10

PR #896, opened 2024-11-20
Clamp drag rectangle to chart area

### LINKED ISSUE #895: Unwanted behavior when dragging outside of chart area
I've noticed a couple of issues with zoom drag rectangles in 2.1.0. (I believe this relates to #811; I was experimenting with that PR's changes locally earlier. I apologize for not following up sooner.)

First, I believe that the behavior is more intuitive / feels cleaner to the end user if the drag rectangle pegs to the edge of the chart area, instead of allowing the rectangle to go beyond the chart area then doing something different with the actual results. (I realize this may be subjective.)

Second, the behavior when the rectangle does go beyond the chart area feels weird. Using https://codepen.io/joshkel/pen/LYwKzwR as a demo, if I click a bit below and to the left of the last point then to the bottom right corner of the canvas (outside of the chart area):

* If I limit the zoom (`limits: { x: { min: "original", max: "original" }, y: { min: "original", max: "original" } }`), then the drag rectangle is shifted (?), so the last point appears even though the rectangle barely excludes it.
* With the default chartjs-plugin-zoom options (no `limits`), I can end up with an X axis value greater than the chart started with. This feels weird for a "zoom in" operation.

https://github.com/user-attachments/assets/24e423b7-ec69-4759-b7ff-dc8c770cc383

joshkel:
I believe this is a better UI, and it also fixes some issues with incorrect zoom ranges when dragging the rectangle outside of the chart area.

To help implement this, I copied the clamp helper function from chartjs-plugin-annotation.

Fixes #895

### DIFF
--- a/src/handlers.js
+++ b/src/handlers.js
@@ -3,6 +3,14 @@ import {zoom, zoomRect} from './core';
 import {callback as call, getRelativePosition, _isPointInArea} from 'chart.js/helpers';
 import {getState} from './state';
 
+/**
+ * @param {number} x
+ * @param {number} from
+ * @param {number} to
+ * @returns {number}
+ */
+const clamp = (x, from, to) => Math.min(to, Math.max(from, x));
+
 function removeHandler(chart, type) {
   const {handlers} = getState(chart);
   const handler = handlers[type];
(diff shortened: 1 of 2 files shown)


kurkle (commented):
A test would be good to prevent regressions in the future 🙂

kurkle (approved):

PR #901, opened 2024-11-22
feat: add trigger to onZoom context

kurkle:
Add `trigger` property to the onZoom context object, indication what triggeder the zoom. Possible values: 'api', 'drag', 'wheel', 'pinch'.


yubiuser:
Thanks.

PR #907, opened 2024-11-27
refactor: convert to typescript

kurkle:
# including some other neat things, like:

- type: module, exports
- prettier
- separate unit tests (just a couple for now)


.eslintrc.yml
```diff
@@ -1,22 +1,36 @@
...
+  jasmine: true
+
+plugins:
+  - "@typescript-eslint"
+  - prettier
+  - html
 
 rules:
+  prettier/prettier: "error"
+  semi: ["error", "never"]
```
LeeLenaleee, line 25: Is there a specific reason you want to turn semi off?
If so, some files still use them like config.ts for vuepress.

I am more of a fan of using them but that can also be something of getting used to

LeeLenaleee (dismissed):

.eslintrc.yml
kurkle: Just a personal preference. Also ok keeping it on.

LeeLenaleee: I guess either way is fine.

Approved the pr

etimberg (dismissed):

PR #913, opened 2024-12-14
Remove unnecessary chart updates on drags

alex-statsig:
Changes a few `chart.update()` calls in the zoom drag handler (i.e. mouse down, mouse move, mouse up) to just use `chart.draw()` instead. The chart update was unnecessary since none of the chart elements had updated. It is also much more expensive to update the chart (incurring costs of things like recomputing the ticks, recomputing any datapoint styling using chart context, etc.) than just to redraw the chart. In some cases with lots of datapoints, this speeds up the function from 100ms -> 1ms. As far as I can tell, the only purpose of the `chart.update()` call was to repaint the shaded area being zoomed, but `chart.draw()` achieves the same purpose.

### DIFF
--- a/src/handlers.ts
+++ b/src/handlers.ts
@@ -45,7 +45,7 @@ export function mouseMove(chart: Chart, event: MouseEvent) {
   if (state.dragStart) {
     state.dragging = true
     state.dragEnd = event
-    chart.draw();
+    chart.draw()
   }
 }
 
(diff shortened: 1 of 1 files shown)


NotTsunami:
@kurkle Would be nice to see this one move along

alex-statsig:
Anything I can do to help get this merged?

kurkle (approved):

alex-statsig:
@kurkle would it be possible to get a new release cut with this?

NotTsunami:
Bump @kurkle

GalexY727:
bump @kurkle 
this is a massive performance improvement.

Issue #238, opened 2019-05-20
Sync pan and zoom across multiple charts


maxxxxxxxim:
This is more of a suggestion. It would be awesome to have the ability to sync pan / zoom across multiple charts! Dygraphs provides this...

benmccann:
If each chart uses the same scale, you could do this by adding an `onPan` and `onZoom` that sets the `min` and `max` for each scale

kylelawson6:
If this method were to be used, how would you get the new scale limits made from the zoom/pan, that would be used to set min/max?

hodbauer:
> UPDATE: ...

Thanks @kylelawson6 this idea is awesome, i tried it and it work like a charm

jledentu:
As @kylelawson6 and @hodbauer explained, I think that this feature can be implemented pretty easily outside of the plugin and shouldn't be added in the core of this plugin (it would add some complexity to handle groups of charts that are synchronized, I guess).
Nevertheless, we may add some documentation (README/wiki) to answer to this question and help future users that want to implement this? What do you think?

tuxd:
I was trying to get this working with https://nagix.github.io/chartjs-plugin-streaming/ but you don't have access to min and max properties. Has anyone got a chart js streaming chart with zoom and pan synced to two charts?

ewmailing:
Can you elaborate on how to do this technique? I tried the following which I found suggested in some other searches that turned up, but hit some bad corner case bugs.

For each of my chart, I implement the onZoomComplete callback. Here is the example for the chart1 callback.

```
	  onZoomComplete: function(the_chart1) {
		  g_chart2.options.scales.x.min = the_chart.chart.options.scales.x.min;
		  g_chart2.options.scales.x.max = the_chart.chart.options.scales.x.max;
		  g_chart2.update();
         }
```

I also implemented a button that will call ```resetZoom()``` for every chart.

This works in the most basic usage, but the bug is as follows:

1. Drag-zoom some area in chart 1. (Both charts will zoom correctly.)
2. Then drag-zoom some area in chart 2. (Both charts will zoom again.)
3. Then hit the reset button to call resetZoom() on both charts.

In step 3, the charts do not reset correctly. It looks like the reset zoomed back to the value in step 2, instead of the original state. As far as I can tell, the original starting state gets lost if you switch between graphs to zoom. I cannot figure out how to get back to the original starting state.

[...]

PR #933, opened 2025-04-02
Extend onZoom and onPan callbacks for syncing multiple charts

vbackeberg:
### Purpose
onZoom and onPan callbacks now return additional data that is useful for syncing multiple charts.

### Changes
- `onZoom` returns the amount it zoomed.
- `onPan` returns the delta it panned.

These values can be used to zoom/pan other charts by the same amount.
The values are basically piped through from the function args, so there is no additional computation.

- `pan()` receives a new PanTrigger parameter.

It can be used in the spirit of ZoomTrigger to distinguish between "api" and user interactions.
This is required to pan other charts when the user pans, but not trigger a loop through the subsequent programmatic calls to pan().

### Example
```
function createChart(chart:Chart, chartElement: HTMLCanvasElement) {
  chart = new Chart(chartElement.getContext('2d')!, {
    type: 'line',
    data: { datasets: items },
    options: {
      plugins: {
        zoom: {
          pan: {
            enabled: true,
            mode: 'x',
            onPan: ({ chart, delta, trigger }) => {
              if (trigger === 'api') return;
              Object.values(Chart.instances).forEach((c) => {
                if (c.id !== chart.id) {
                  c.pan({ x: delta.x });
                }
              });
            }
          },
          zoom: {
            wheel: { enabled: true },
            mode: 'xy',
            onZoom: ({ chart, amount, trigger }) => {
              if (trigger === 'api' || amount === undefined) return;
              Object.values(Chart.instances).forEach((c) => {
                if (c.id !== chart.id) {
                  c.zoom(amount);
                }
              });
            }
          },
        },
      },
    }
  });
}
```

### Background
So far there is no good way to sync zoom and pan across multiple charts that I could come up with or have found during my research.

The only approach I found was this from 2019:
https://github.com/chartjs/chartjs-plugin-zoom/issues/238#issuecomment-2002819777

The changes I propose solve the issue of syncing multiple charts in a simple way while not breaking anything.
I think it would be a good addition to the library that could help many others.

### DIFF
--- a/src/core.ts
+++ b/src/core.ts
@@ -3,7 +3,7 @@ import { panFunctions, updateRange, zoomFunctions, zoomRectFunctions } from './s
 import { getState, type OriginalScaleLimits, type ScaleRange, type State, type UpdatedScaleLimits } from './state.js'
 import { directionEnabled, getEnabledScalesByPoint } from './utils.js'
 import type { Chart, Point, Scale, UpdateMode } from 'chart.js'
-import type { LimitOptions, ZoomTrigger } from './options.js'
+import type { LimitOptions, PanTrigger, ZoomTrigger } from './options.js'
 import type { ZoomAmount } from './types.js'
 
 function shouldUpdateScaleLimits(
@@ -89,7 +89,7 @@ export function zoom(chart: Chart, amount: ZoomAmount, transition: UpdateMode = 
 
   chart.update(transition)
 
-  zoomOptions?.onZoom?.({ chart, trigger })
+  zoomOptions?.onZoom?.({ chart, trigger, amount: { x, y, focalPoint } })
 }
 
 export function zoomRect(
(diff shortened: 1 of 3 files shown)


stsrki:
I have just noticed this PR, and it is just the feature that is perfect for our use case, too. https://github.com/Megabit/Blazorise/pull/6109

Questions for maintainers. Is there any chance this could be merged soon? Thanks.

vbackeberg:
I simplified the amount return value a bit, since there was some obsolete actualAmount calculation before, and squashed the commits. Now it is pretty plain simple.

We are using it and so far works very well. :)

devs, let me know if I need to do anything.

stsrki:
@kurkle, apologies for the direct mention. I just wanted to check if this PR looks good to merge.

Thanks!

technyon:
Hi. Any update on this? Merging this would be very appreciated.

technyon:
Hi,

we'd really need this functionality. Would it be possible to review and merge it? Looking at the diff, it's only 8 lines that have been updated.

kurkle:
@vbackeberg would it be possible for you to still continue with this? (needs a rebase at least)

PR #949, opened 2026-07-25
fix: pan category scales whose categories are narrower than a pixel

### LINKED ISSUE #946: With ≥ 2× more categories than pixels, panning jumps straight to the first/last category
With `pan: { enabled: true }` on a category axis, if the chart has at over twice as many visible categories as its width in pixels, a tiny drag to the left or right snaps the visible range all the way to the first or last category, with no in-between.

This is due to incorrect rounding in `panCategoryScale`:

https://github.com/chartjs/chartjs-plugin-zoom/blob/4aa6d11aed1c271bc8698bd414adc9587e1bfbb9/src/scale.types.js#L241-L245

When `scaleLength / range < 0.5`, `stepDelta` rounds to `0`, and `stepSize` becomes `Infinity`.

There’s no reason to round `stepDelta` to an integer. It should be kept fractional to accurately reflect the fractional size of each category, which is significant when that size is small.

### Reproduction

https://jsfiddle.net/anderskaseorg/vkcn85mp/

```html
<canvas id="ctx">
<script type="module">
import { BarController, BarElement, CategoryScale, LinearScale, Chart } from "https://esm.sh/chart.js@4.5.1";
import zoomPlugin from "https://esm.sh/chartjs-plugin-zoom@2.2.0";
Chart.register(BarController, BarElement, CategoryScale, LinearScale, zoomPlugin);
const chart = new Chart(ctx, {
  type: "bar",
  data: {
    datasets: [{ data: [...Array(10000).keys()] }],
    labels: [...Array(10000).keys()],
  },
  options: { plugins: { zoom: { pan: { enabled: true } } } },
});
chart.zoom(1.25);
</script>
```

rajanpanth:
## Summary

Fix category-scale panning when the chart displays more than twice as many categories as it has pixels.

`panCategoryScale()` rounds the number of pixels per category before calculating how many categories a drag should move. When categories are narrower than half a pixel, that rounded value is zero, so the calculated step count becomes infinite and any drag snaps directly to the first or last category.

## Changes

- keep the fractional pixels-per-category value only when rounding would produce zero
- preserve the existing rounded behavior for categories that are at least half a pixel wide
- add regression coverage for proportional sub-pixel panning and both category-range boundaries
- retain coverage proving ordinary multi-pixel category panning is unchanged

## Testing

- `npm run test-unit` — 20 specs, 0 failures
- `npm run typecheck` — passed
- `npm run test-types` — passed
- `npm run lint` — 0 errors and 15 pre-existing warnings, matching `master`
- `npm run build` — passed, including declaration generation
- `git show --check HEAD` — passed

The browser Karma suite was not rerun for submission. Earlier Windows runs were nondeterministic across both this branch and unmodified `master`; the consistent logarithmic-wheel failure also reproduces on `master`. The changed category-pan calculation is a mathematical no-op for every existing fixture where rounded pixels per category is nonzero.

## Compatibility

No public API or type changes. Existing category-pan behavior is unchanged unless the prior calculation rounded the step size to zero.

Fixes #946


rajanpanth:
Friendly follow-up on this PR 👋

When you have a moment, could you please take a look? I’m happy to rebase, update tests, or make any requested changes.

Thanks!

src/scale.types.ts
```diff
@@ -219,7 +219,11 @@ function panCategoryScale(scale: Scale, delta: number, limits: LimitOptions) {
   // The visible range. Ticks can be skipped, and thus not reliable.
   const range = Math.max(max - min, 1)
   // How many pixels of delta is required before making a step. stepSize, but limited to max 1/10 of the scale length.
-  const stepDelta = Math.round(scaleLength(scale) / Math.max(range, 10))
+  // Categories can be narrower than a pixel, in which case rounding the step down to zero would
+  // make stepSize infinite and send every pan straight to the first or last category, so keep the
+  // fractional step for that case.
+  const pixelsPerStep = scaleLength(scale) / Math.max(range, 10)
+  const stepDelta = Math.round(pixelsPerStep) || pixelsPerStep
```
kurkle, line 226: I think removing the round completely would make more sense. But that would probably mean updating couple of fixtures. Would you be willing to do that?

kurkle: actually, if you are, do that in a followup :)

kurkle (approved):

Issue #938, opened 2025-06-25
Find a solution for hammerjs


micheljung:
hammerjs [has been unmaintained for 6 years](https://github.com/hammerjs/hammer.js/graphs/contributors). It's last version has been released 9 years ago.

Since we updated to chart.js 4, we get this:

    ▲ [WARNING] Module 'hammerjs' used by 'node_modules/chartjs-plugin-zoom/dist/chartjs-plugin-zoom.esm.js' is not ESM

It seems as if it's only a matter of time until things start to fall apart because of hammerjs - so maybe it's time to find an alternative?

hesprs:
I made an alternative for Hammer.js called [Pointeract](https://pointeract.consensia.cc). It's TypeScript-native, tiny (2.7KB gzipped and can be even smaller if you use less modules), and handles pan and zoom well.

It currently does not have many stars since it's new. Here's an [interactive demo](https://pointeract.consensia.cc).

Hope can help you.

PR #950, opened 2026-09-09
Replace Hammer.js with native Pointer Events

micheljung:
> **Disclosure:** This change was authored by Claude (Anthropic's Claude Code AI assistant) and reviewed by @micheljung.

## Why
`hammerjs` (last released 2016) was a required dependency used only for touch pinch/pan gesture recognition. This replaces it with the browser's native **Pointer Events**, leaving the plugin with **zero runtime dependencies**.

## What changed
- **`src/gestures.ts`** (new) replaces `src/hammer.ts`: a `GestureManager` tracks native pointers and reproduces the previous pan/pinch behaviour — pan threshold, mouse modifier-key gate, pinch focal point, cumulative pinch scale, and the pinch→pan finger-lift transition. It sets `touch-action: none` while active (as Hammer did) and cleans up on `stop`. A `GestureEvent` type replaces the `HammerInput` used in the public option types (`onPanStart`/`onPanRejected`/…).
- **`plugin.ts`**: dropped the `import Hammer` feature-detect guard; gesture setup/teardown is now unconditional.
- **Dependencies**: removed `hammerjs` + `@types/hammerjs` (runtime) and `hammer-simulator` + `ng-hammerjs` (test/docs helpers). Refreshed the build/test toolchain (rollup, swc, `@rollup/*`, c8, eslint plugins, karma helpers).
- **Tests**: karma pinch/pan specs are driven with synthetic `PointerEvent`s (`jasmine.simulatePan` / `jasmine.simulatePinch`) instead of the Hammer simulator.
- **Docs/sample**: removed the `hammerjs` script tag and the vuepress `ng-hammerjs` alias.

## Deliberately held back
`chart.js` (dev pin) stays on the 4.3.x line — 4.4+ tightens `Point` nullability, which surfaces pre-existing strict-null issues in core files unrelated to this change. `typescript` stays on 5.0 because `typedoc@0.23` caps it at `<5.1`. The **`chart.js` peer range is unchanged (`>=3.2.0`)**.

## Verification
`build`, `typecheck`, `test-types`, `lint`, and `test-unit` pass. The karma gesture specs need a browser/CI to run and were not executed in the authoring environment — please let CI validate them.


micheljung:
Fixes #938

src/gestures.ts
```diff
@@ -0,0 +1,364 @@
...
+      scale,
+      pointers,
+      target: this.canvas,
+      srcEvent,
+      pointerType: srcEvent.pointerType,
+    }
+  }
+
+  private onPointerDown = (event: PointerEvent) => {
+    this.pointers.set(event.pointerId, {
```
kurkle, line 241: suggestion: I think this should be limited to button 0, right?

```suggestion
    if (event.pointerType === 'mouse' && event.button !== 0) return

    this.pointers.set(event.pointerId, {
```

src/gestures.ts
```diff
@@ -0,0 +1,364 @@
...
+      clientX: event.clientX,
+      clientY: event.clientY,
+      pointerType: event.pointerType,
+    })
+
+    const { pan: panOptions, zoom: zoomOptions } = this.state.options
+
+    if (this.pointers.size >= 2 && zoomOptions?.pinch?.enabled) {
```
kurkle, line 249: An accidental third finger would restart the pinch. Probably safe to ignore any extra pointers.

```suggestion
    if (this.pointers.size === 2 && zoomOptions?.pinch?.enabled) {
```

src/gestures.ts
```diff
@@ -0,0 +1,364 @@
...
+      // A second finger starts a pinch; a pan already in progress ends first.
```
kurkle, line 250: if pinch is not enabled, this block would not be hit. A second finger down while panning and then lifting the second finger would cause a jump. 

So I'd move this up and change the if to `if (this.panning && this.pointers.size > 1)`

src/gestures.ts
```diff
@@ -0,0 +1,364 @@
...
+      if (this.panning) {
+        endPan(this.chart, this.state)
+        this.panning = false
```
kurkle, line 253: this.panning and state.panning are tracking the same thing, right? no need for both

src/gestures.ts
```diff
@@ -0,0 +1,364 @@
...
+      srcEvent,
+      pointerType: srcEvent.pointerType,
+    }
+  }
+
+  private onPointerDown = (event: PointerEvent) => {
+    this.pointers.set(event.pointerId, {
+      clientX: event.clientX,
+      clientY: event.clientY,
+      pointerType: event.pointerType,
```
kurkle, line 244: the pointerType is never used, is it needed?

kurkle (commented):
Nice work, thank you 👍 

I made couple of comments, and additionally the package changes, other than dropping hammer should be left out.

src/gestures.ts
micheljung: Done.

src/gestures.ts
micheljung: Done, went a little further: pointers past the second are not tracked at all. With just `=== 2` the extra finger still skews the centroid mid-pinch, and lifting it ends the pinch while two fingers are still down.

src/gestures.ts
micheljung: Done. Needed one more piece: pointerup now re-seeds the pan start when a single pointer remains, otherwise lifting the *pan* finger and keeping the other one jumps by the distance between them.

src/gestures.ts
micheljung: Dropped it, using `state.delta` instead. Not quite the same thing: `state.panning` is only set once `handlePan` runs, so with `onPanStart` returning false a plain swap re-fires the callback on every move. The gesture now stops offering the pan after a rejection.

src/gestures.ts
micheljung: Removed 👌

micheljung:
Also reverted the unrelated dependency bumps — `package.json` is down to the hammer removals only.

kurkle:
Conflicts need to be resolved. Also the text select that hammer did, sounds like sonething worth doing

kurkle (approved):

Issue #947, opened 2026-06-16
Wheel zoom uncontrollable on high-resolution touchpad: fixed per-event speed, laggy synchronous redraw for each event


andersk:
Two problems with the `wheel` event handler compound to make zooming completely uncontrollable on a touchpad with high-resolution scrolling, which emits `wheel` events with small `deltaY` at a high frequency:

1. The handler ignores the event’s [`deltaMode`](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaMode) and the magnitude of [`deltaY`](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaY): it only takes the sign into account. So the high event frequency leads to much more zooming than the user intended.

https://github.com/chartjs/chartjs-plugin-zoom/blob/4aa6d11aed1c271bc8698bd414adc9587e1bfbb9/src/handlers.js#L246-L247

2. The handler redraws the chart synchronously for each event, rather coalescing multiple events across a [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame) callback. If the redraw is even just a little slower than the time between events, a backlog of events builds up in the queue, and the chart lags farther and farther behind real time.

https://github.com/chartjs/chartjs-plugin-zoom/blob/4aa6d11aed1c271bc8698bd414adc9587e1bfbb9/src/handlers.js#L257

### Reproduction

https://jsfiddle.net/anderskaseorg/ezy5guf2/

```html
<canvas id="ctx">
<script type="module">
import { Chart, LinearScale, PointElement, ScatterController } from "https://esm.sh/chart.js@4.5.1";
import zoomPlugin from "https://esm.sh/chartjs-plugin-zoom@2.2.0";
Chart.register(LinearScale, PointElement, ScatterController, zoomPlugin);
const chart = new Chart(ctx, {
  type: "scatter",
  data: {
    datasets: [{
      data: [...Array(10000).keys()].map(() => ({ x: Math.random(), y: Math.random() })),
    }],
  },
  options: { plugins: { zoom: { zoom: { wheel: { enabled: true } } } } },
});
</script>
```

PR #951, opened 2026-09-10
fix: scale wheel zoom by the scrolled distance

Pareder:
The wheel handler only looked at the sign of `deltaY`, so a high resolution trackpad - small deltas at a high frequency - zoomed a full notch per event. The zoom is now proportional to the distance scrolled, measured per `deltaMode` and capped at one notch per event.

Addresses https://github.com/chartjs/chartjs-plugin-zoom/issues/947


kurkle:
Thank you for the PR. I'm afraid this is not that simple, because browsers behave differently in different operating systems too. There is a good article about this stuff: https://danburzo.ro/dom-gestures/

I'd suggest the easy compromise there (for now at least):

>/*
>	Note: we're using `Math.sign()` and `Math.min()` 
>	to impose a maximum on the *absolute* value 
>	of a possibly-negative number.
> */
>dy = Math.sign(dy) * Math.min(24, Math.abs(dy));

That should make this usable, what do you think?

andersk:
Note that this does not address the part of #947 about event queue backlogging due to synchronous redraws.

Pareder:
@kurkle Applied your proposition, works good for me.

@andersk is right, this PR addresses the first part of his issue, that's why I used non-closing keyword.

kurkle (approved):

PR #958, opened 2026-09-11
fix: never round fractional category size for panning

andersk:
- Followup to https://github.com/chartjs/chartjs-plugin-zoom/pull/949#discussion_r3982303877


kurkle:
The red CI is why I suggested this to be a follow up 😈 
You'll need to change the tests a little.

kurkle (approved):

PR #972, opened 2026-09-17
Docs vitepress

kurkle:


kurkle:
@etimberg I was thinking something like this could be the way out from vuepress for chart.js. WDYT?

etimberg (approved):

